PR #173, opened 2012-07-29
Switch signup email input to use HTML5 email type to save mobile browsers a tap or two.

mikelietz:
I'm not entirely certain I did the SCSS stuff right - but I was trying to preserve what the input looks like by adding the new type to the same rule it used before.


mikelietz:
There are other text inputs. I'll do one request for all of them.

PR #174, opened 2012-07-29
Use HTML5 email inputs for emails to save mobile browsers a couple taps.

mikelietz:
I'm not sure about the SCSS change - the idea was to use the same rule as the existing text input so that the forms look the same.

### DIFF
--- a/bookie/static/css/responsive.scss
+++ b/bookie/static/css/responsive.scss
@@ -190,7 +190,7 @@
             }
         }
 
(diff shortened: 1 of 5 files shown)


mitechie:
Thanks for the pull request! There was one change needed in the base.scss that needed to be updated for desktop viewers as well I added and rebased up. I've added you to the credits.txt, feel free to update that to the full name if you want.

Merged and deployed. Might need to clear the cache on the mobile machine for the css updated. It's not versioned like JS currently is.

PR #185, opened 2012-08-25
Start celery with « make run » command

harobed:
Start celery with « make run » command without it mails are neved send and user can't create new user to test Bookie.

Best regards,
Stephane

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,9 @@ ropevim.vim
 *.pid
 /download-cache
 /extensions/chrome_ext/lib
+bookie.ini
+bookie/static/
+bookie_index/
+celerybeat-schedule
+docs/_build/
+share/
--- a/Makefile
+++ b/Makefile
@@ -298,7 +298,7 @@ clean_chrome:
 	fi
 
 
-run: run_combo run_app
+run: run_celery run_combo run_app 
 run_celery:
 	BOOKIE_INI=$(BOOKIE_INI) $(CELERY) --pidfile celeryd.pid &
 run_dev: run run_css autojsbuild
@@ -313,7 +313,7 @@ run_livereload:
 autojsbuild:
 	$(PY) scripts/js/autojsbuild.py -w $(BOOKIE_JS) -b $(JS_BUILD_PATH)/b
 
-stop: stop_combo stop_app
+stop: stop_combo stop_app stop_celery
 stop_dev: stop stop_css
 stop_celery:
 	kill -9 `cat celeryd.pid` || true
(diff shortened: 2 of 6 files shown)


mitechie:
Thanks a ton for the pull request!

The make run makes sense as it's required to run things. I've added a run_webapp for my old purpose of just working on the web ui without dealing with the running celery backend. 

I've pushed that change up to the develop branch and added you to the CREDITS.txt.

Now for the other change, it's my goal to deprecate the whole usermgr and move its work to the bookie api python package.

https://github.com/mitechie/bookie_api

It currently will get a list of users and it'd be great to make the password reset work part of that. The big problem is that there's not currently an admin-only api call to force a password reset. It's marked as @todo in the api docs atm. 

I'm not going to merge that in currently, but would love to work together on adding in the admin api calls to the web app and building in support for them into the bookie python client instead as a central place to keep all misc work like this going forward.

harobed:
> Now for the other change, it's my goal to deprecate the whole usermgr and move its 
> work to the bookie api python package.
> 
> https://github.com/mitechie/bookie_api

Ok I understand.

> I'm not going to merge that in currently, but would love to work together on adding in the admin api
> calls to the web app and building in support for them into the bookie python client instead as a 
> central place to keep all misc work like this going forward.

Ok but for the moment, can you append this feature ?
It's useful for Bookie user who want install personal Bookmark on their website (it's my goal).

After, we'll can change this bookie-manage with API implementation ?

Personally, now, before append new feature, I think the first goal of Bookie is to simplify installation… management tools to make it popular and get more contributor like me.

What are your comments ?

mitechie:
Thanks for your understanding. 

Did you see the current code in scripts/admin/usermgr.py ? That allows for adding users currently. This is why if we reworked this code I'd prefer the work went into the api/cli client vs more direct user scripts.

```
% ./bin/python scripts/admin/usermgr.py --help
usage: usermgr.py [-h] [--list] [--ini INI] [--new] [--username USERNAME]
                  [--email EMAIL] [--reset-password] [--password PASSWORD]

Update existing bookbmarks with the readability parsed

optional arguments:
  -h, --help            show this help message and exit
  --list, -l            Get the list of users from the system.
  --ini INI             The .ini configuration file to use for getting db
                        access
  --new, -n             Add new user to the system
  --username USERNAME, -u USERNAME
                        The username to use for this task.
  --email EMAIL, -e EMAIL
                        The email to use for this task.
  --reset-password, -r  Reset the user's password
  --password PASSWORD, -p PASSWORD
                        The password to use for this task.
```

harobed:
I see it, I used this code to create bookie-manage script.

I think modestly that bookie-manage code is "better", it use sub command, I have appended documentation page…

mitechie:
For the record, the bookie-api client 0.4.2 adds in the user management commands. It allows for adding, listing, and simple deletion cases from the admin only api and the command line client. Let me know if you hit any issues using it.

harobed:
I'll try to found the time to test it this week.

PR #191, opened 2012-10-16
Add RSS feature

ilesinge:
Added RSS 2.0 support:
- 4 new routes, corresponding to 'recent' routes (global RSS, global filtered by tag, per user, per user filtered by tag).
- Appropriate links added to the head of 'recent' views

What is a bookmarking service without RSS? :)

### DIFF
--- a/bookie/routes.py
+++ b/bookie/routes.py
@@ -27,11 +27,17 @@ def build_routes(config):
     config.add_route("bmark_recent", "recent")
     config.add_route("bmark_recent_tags", "recent/*tags")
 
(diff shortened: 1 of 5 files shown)


mitechie:
I've merged this in and deployed with some tweaks on my end. I've started a follow up branch that will work on updating the client side JS to update the rss feed urls as the changes to the visible data occur. It's not going to be ready for a bit yet and I wanted to get this out. 

I've added ilesinge to the CREDITS.txt and did some small tweaks, added some tests, etc.

PR #241, opened 2013-08-31
extended make user

drewverlee:


mitechie:
I went to look at merging this in but it's got some issues. 

There are several .txt files that should not be committed to the tree. There's also some spacing issues. Run `make lint` and check out some of the too many line errors. There's already a base of errors though, so don't try to remove them all.

drewverlee:
Thanks,

Sorry for the late reply. I think i cleaned up most of those issues. I'll
try to push the changes next CHC. I'm going to use every CHC's to try to
work on bookie in some way. I'm curious if you would want to devote
sometime to explaining more about bookie so that I could be more effective
at helping with it. I understand that's a time commitment on your part but
i guess i can promise at least 2 hours a week to work on bookie for roughly
2 semesters.

I can't imagine finding an opensource project that will be easier to work
with given your accessibility.

mitechie:
Sure thing, I'm not going to be at CHC this week, but happy to trade some mentoring patches.

drewverlee:
Great, let me know what times would work for you. I'm more or less free any
time except Monday and Wednesday 1:00 - 8:00. obviously on W im at CHC's.

I tried to create some tests but i'm not unsure of the 'big picture' to get
anything moving. I spent a bit of this CHC's reading about pyramid.

PR #247, opened 2013-10-04
Allow non-interactive `make sysdeps`

fitztrev:
I was working on a Vagrant script that would fire up and provision a VM automatically so I could run Bookie. This change will allow my provisioner to install the system dependencies without prompting me to confirm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,11 @@ clean_all: clean_venv clean_js clean_css clean_chrome clean_downloadcache
 
 .PHONY: sysdeps
 sysdeps:
-	sudo apt-get install $(SYSDEPS)
+	if [ $(NONINTERACTIVE) ]; then \
+		sudo apt-get install -y $(SYSDEPS); \
+	else \
+		sudo apt-get install $(SYSDEPS); \
+	fi
 
 .PHONY: install
 install: $(BOOKIE_INI) all first_bookmark css


mitechie:
This is very cool and a good idea. The way this works you need to set the NONINTERACTIVE=y to be correct. 

I'm thinking a better long term approach may be to have that be an overall setting to all make commands. So the idea would be to say "Do this make command in a non-interactive fashion" and any task can check for that environment variable. 

So I'd suggest we make NONINTERACTIVE a full env var we support. 

NONINTERACTIVE=true make sysdeps && make install

The check would then be more based off of 
http://stackoverflow.com/questions/307503/whats-the-best-way-to-check-that-environment-variables-are-set-in-unix-shellscr

So rather than adding the var to the makefile, the task would just do a if/else check if that exists. 

if [ "$NONINTERACTIVE" ] ; then 
    apt-get install -y .....;
 else
    apt-get install 
fi

fitztrev:
I agree. I like that approach. I've updated the pull request to make NONINTERACTIVE an environment variable instead.

New method:

```
$ NONINTERACTIVE=true make sysdeps
```

or

```
$ export NONINTERACTIVE=true
$ make sysdeps
```

mitechie:
Still needs docs and check credits file for Trev's name there. With this could re-open.

PR #250, opened 2013-10-16
Mark anonymous endpoints in API docs

fitztrev:
I noticed some of the endpoints allow anonymous access with no API key required.

```
/bmarks
/:username/bmarks
/bmarks/popular​
/:username/bmarks/popular
/bmarks/search
/:username/bmarks/search
/:username/tags/complete
```

Also, `/bmarks/popular​` was using the same description and example as `/:username/bmarks/popular`.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -203,7 +203,7 @@ Usage
 
 Return a list of the most recent bookmarks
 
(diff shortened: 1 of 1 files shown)


mitechie:
Thanks! <3 doc updates!

PR #258, opened 2013-12-01
Handle exception when adding a bmark w/o url

fitztrev:
When doing an empty POST to /:username/bmark, Pyramid would throw an
exception ("No JSON object could be decoded").

Also, updated api docs to include required `url` param.

### DIFF
--- a/bookie/views/api.py
+++ b/bookie/views/api.py
@@ -173,7 +173,7 @@ def bmark_add(request):
             params = request.json_body
         else:
             raise ValueError('No url provided')
-    except ValueError, exc:
+    except ValueError:
         request.response.status_int = 400
         return _api_response(request, {
             'error': 'Bad Request: No url provided'
--- a/bookie/views/api.py
+++ b/bookie/views/api.py
@@ -166,15 +166,17 @@ def _update_mark(mark, params):
 def bmark_add(request):
     """Add a new bookmark to the system"""
     rdict = request.matchdict
-
-    if 'url' in request.params or 'hash_id' in request.params:
-        params = request.params
-    elif 'url' in request.json_body or 'hash_id' in request.json_body:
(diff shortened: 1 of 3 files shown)


bookie/views/api.py
```diff
@@ -155,11 +155,14 @@ def _update_mark(mark, params):
...
-    elif 'url' in request.json_body or 'hash_id' in request.json_body:
-        params = request.json_body
-    else:
+    try:
+        if 'url' in request.params or 'hash_id' in request.params:
+            params = request.params
+        elif 'url' in request.json_body or 'hash_id' in request.json_body:
+            params = request.json_body
+        else:
+            raise
```
mitechie: I'd prefer to catch the specific exception when we can. It looks like this exception should be a case of a ValueErorr thrown from the json module. It comes from running json.loads("") on an empty string, which is the body in the empty post case.

So a change to try: except ValueError: and then the else block can raise a ValueError('No url provided') or something to hint at why the else clause is a 400.

mitechie:
Thanks for this change. I appreciate things that help make the api more robust to user requests.

I'd love to see a test case for this added that POSTs without the fields and verifies it gets back the JSON {error: } response. If you'd like I can help with the test part. Let me know. 

My only other feedback is that we should catch/throw specific exceptions whenever possible. I think working with the original ValueError exception would be worthwhile.

fitztrev:
I added the code to catch the specific exception. I also added another test for when there does exist a POST body, but no `url` in the JSON.

Updated and rebased. Let me know what you think.

fitztrev:
Tested, linted, rebased, pushed, and hopefully ready to go. :smile:

mitechie:
Thanks so much for going through this process and adding the tests!

PR #259, opened 2013-12-01
Add jsonp support to API calls

fitztrev:
Adds jsonp renderer to pyramid config
Update api docs w/ optional callback param


mitechie:
Thanks for this change. Since it's just using the built in pyramid tooling not looking for a specific test case around the callback parameter. Thanks for update the docs with this!

PR #262, opened 2013-12-05
create factory function make import queue

drewverlee:
A factory function for making a import queue.


mitechie:
Factory helper looks good, let's use it!

PR #263, opened 2013-12-06
Allow anonymous API requests for bmark_get

fitztrev:
Currently, GET /:username/bmark/:hash_id only allows you to access detail about your own bookmarks. I'd like to open it up so apps can get it, too.

### DIFF
--- a/bookie/views/api.py
+++ b/bookie/views/api.py
@@ -91,7 +91,7 @@ def ping_missing_api(request):
 
 
 @view_config(route_name="api_bmark_hash", renderer="jsonp")
-@api_auth('api_key', UserMgr.get)
+@api_auth('api_key', UserMgr.get, anon=True)
 def bmark_get(request):
     """Return a bookmark requested via hash_id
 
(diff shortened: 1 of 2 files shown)


bookie/views/api.py
```diff
@@ -103,7 +103,11 @@ def bmark_get(request):
     params = request.params
 
     hash_id = rdict.get('hash_id', None)
-    username = request.user.username
+
+    if request.user and request.user.username:
```
mitechie, line 107: I don't think this is correct for what we want to do. The username is available in the matchdict in the case that we're not defaulting to the user that is logged in. The api url would be /username/bmark/hash_id I believe.

So I'd say that we just respect the username from the request.matchdict (rdict in this case)

I don't think it's ever valid to get a bookmark based on hash_id alone as multiple people are able to bookmark a single url.

mitechie:
Thanks for this change. I think it's a good one to make and a start at not requiring auth for requests that don't really require it. 

I've commented on the change specifically, I'd also see if you could, and I'd be happy to help with, add some tests for the changes. Based on the comment I added, I think there's a few we need.
- Can I get a bookmark without being logged in
- Can I make sure to get the correct bookmark for two users that saved the same url (common hash_id)
- If I do attempt to get a bookmark with an api_key, that it works sanely.

That last one is the reason that this all started out only getting your bookmark with an api key. The issue is that the api assumes that when you get a url with username in the matchdict and an api key, that it attempts to authorize you as that user. If you attempt to get my bookmark [/admin/bmark/123456] and you supply an api key, there's no place for you to supply the username that api key is good for. 

I think the reasonable thing to do is just ignore the api_key. Otherwise, it'll break things.

mitechie:
Merged with updates. Thanks for the changes and helping to improve the API!

PR #272, opened 2014-01-30
Support search with username filter.

mitechie:
- Add an input on the search form to limit search to only your bookmarks
- Update the API to include the username as part of the search
- Make sure that it checks both in the GET, in the route (mdict)


mitechie:
Searching needs to be redone, but will run with this as a temp fix for things.

PR #280, opened 2014-02-25
issue278

preddy5:
@mitechie could you review the code and comment on it.

Thank you
Pradyumna


mitechie:
Thanks for this update. This seems like a generally good idea but ideally we'd have this tested.

It brings out a larger issue that the edit Javascript isn't wrapped in a Y.View that we can more easily test. It's just raw JS in the template which isn't ideal.

I'd like to hold onto this until we can get the JS in this template moved into the bookie/static/views.js as a new Javascript view and then add tests for it to verify it's working and doesn't break in the future.

If you're up to trying that out I'd be happy to help or I'll try to get to the code cleanup later this week and then see if we can adjust this pull request to fit the new code layout.

preddy5:
@mitechie thanks for the comment, I will edit and pull the patch in some time.

sambuddhabasu:
with the current changes, if we type http://www.google.com, we get http://http://www.google.com
this needs to be fixed. also i am working right now on moving the function and then writing tests for it

carpedm20:
So, I suggested that "http://httsp://" this kind of problems can be avoided by using "onblur" event rather than "onkeyup" event in  #285.

sambuddhabasu:
@carpedm20 you are right. the 'onblur' event will make more sense than the 'onkeyup' event

preddy5:
@mitechie could you comment on the code, I will pull the tests in next commit

bookie/static/js/bookie/view.js
```diff
@@ -917,7 +917,55 @@ YUI.add('bookie-view', function (Y) {
...
+    ns.BmarkEditView = Y.Base.create('bmark-edit-view', Y.View, [], {
+        
+        /**
+         * Binds event to html view.
+         *
+         * @method _bind
+         *
+         */
+        _bind: function () {
+            Y.one('#url')._node.onblur = this._check_and_edit_url;
```
mitechie: this is great. The event should be able to be in a YUI View events {} object and then we can skip the _bind method for this bit of work. See http://yuilibrary.com/yui/docs/view/#handling-dom-events

bookie/templates/bmark/edit.mako
```diff
@@ -103,7 +103,7 @@
                'with_submit': false
             });
             tagcontrol.render();
-
+            bmark_edit_view = new Y.bookie.BmarkEditView();
```
mitechie, line 106: you have to call render on the View even thought it's not doing much for the view to bind events. 

So a simple addition of bmark_edit_view.render() should work.

preddy5, line 106: I dont think eventlistener is not binded because of not calling render(Eg: render function of accountview is not called in accounts/index.mako), the event is not getting binded even if i am using "events" from the  above link

mitechie, line 106: RIght, but what I mean is that you should be able to add an events object

```
events: {
    '#url': {
        'click': '_check_and_edit_url'
    }
},
```

Then when you create the view, call .render() on it and it will bind that event and you can drop the _bind() function.

mitechie:
Thanks! This looks really good and I look forward to seeing the commit that adds the test to the validation function.

bookie/static/js/tests/test_view.js
```diff
@@ -292,6 +292,28 @@ YUI.add('bookie-test-view', function (Y) {
...
+        },
+
+        tearDown: function () {
+
+        },
+
+        test_Url_edit: function () {
+            var page = new Y.bookie.BmarkEditView();
+            page.render();
+            Y.one('#url')._node.value = 'google.com';
```
mitechie: You should be able to just skip the _node and do Y.one('#url').set('value', 'google.com');

bookie/static/js/tests/test_view.js
```diff
@@ -292,6 +292,28 @@ YUI.add('bookie-test-view', function (Y) {
...
+            Y.one('#url').simulate('blur');
+            editedurl = Y.one('#url')._node.value;
```
mitechie: same here to get at the value then.

bookie/static/js/bookie/view.js
```diff
@@ -917,7 +917,45 @@ YUI.add('bookie-view', function (Y) {
...
 
+        /**
+         * General initializer
+         *
+         * @method initializer
+         * @param none
+         *
+         */
+        initializer: function () {
+            Y.one('#url')._node.onblur = this._check_and_edit_url;
```
mitechie: this should Y.one('#url').on('blue, this._check_and_edit_url); _node is private and should not be accessed.

mitechie:
Thanks! The change is looking really good and adding tests is awesome. There's a small tweak to not access _node and I'll pull it down to test out and land.

mitechie:
Thanks for this update. I've landed this change into the main develop branch.

Issue #283, opened 2014-02-25
Loop on last page of documentation


aniruddhkanojia:
There is a loop on the last page of documentation. The 'previous' link redirects to this page again.

"http://docs.bmark.us/en/latest/events.html"

PR #284, opened 2014-02-25
Patch for issue #283

aniruddhkanojia:


mitechie:
Hmm, won't this remove the events page from the table of contents entirely?

aniruddhkanojia:
No it won't be removed completely as events is listed once again in the doctree of development.rst

mitechie:
When I test this the Events items is not in the table of contents and there's no way to get to the events page. 

I end up just having

```
 * Installing Bookie                                                        
 * User Docs                                                                
 * Browser Extensions                                                       
 * Hosting your Bookie installation                                         
 * Development 
```

aniruddhkanojia:
Thanks for the response .It was my bad , I didn't see the table of contents but  I think you can still reach events using the next link (after a couple of clicks ) , also you should be able to reach events page using the links in the text.

To have both table of contents and no loop , removing events from toctree of development.rst  and retaining it in index.tst will do the trick.

Issue #278, opened 2014-02-24
URL requires user to type "http://"


sanchitgangwar:
While adding a new bookmark using the web interface, the URL field requires the user to type "http://" manually.

I think that the assumption here is that most users would copy and paste the URL, however if they choose to enter it manually it'd definitely be a pain to type "http://" or "https://" manually.

mitechie:
Thanks for the issue. Good call. Marking this as a good bite size issue.

PR #285, opened 2014-02-25
Another patch for issue #278

carpedm20:
I think onblur event might be a better solution to check url. Because if I accidentally put a 's' at wrong position, "http://httsp://" like things will be made. Please review.


mitechie:
This is a very good point! I think this is best done as a code review comment on the original pull request. I'm going to close this one and look forward to moving forward in the first pull request.

Issue #279, opened 2014-02-24
Problem in installation of bookie


aka001:
In the command:
make sysdeps
If you have an old virtual machine or the folder is in a folder which has a space in between it like "Ubuntu One" then it fails to download all packages.

aka001:
The old virtual machine problem can be resolved by installing a new one.
And the directory can be renamed to a non-space directory. After this follow the whole process again(starting from cloning the repository).
A fix for the spaced directory is needed however.

mitechie:
Thanks for the report. You're right that a space in the path will cause issues. The Makefile needs to be updated to quote paths to avoid this.

aka001:
Okay. :)

PR #286, opened 2014-02-25
Patch for issue #279

aka001:
I have quoted the directory path so that no further problem will be shown while running the command for directories containing spaces.
It relates to issue #279 9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ deps: venv
 	fi
 	@echo "Making sure the latest version of pip is available"
 	# $(PIP) install -U pip
-	$(PIP) install --no-index --no-dependencies --find-links file:///$(CACHE)/python -r requirements.txt
+	$(PIP) install --no-index --no-dependencies --find-links file:///"$(CACHE)"/python -r requirements.txt
 
 # TESTS
 #
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ deps: venv
 	fi
 	@echo "Making sure the latest version of pip is available"
 	# $(PIP) install -U pip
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -135,7 +135,7 @@ deps: venv
 	fi
 	@echo "Making sure the latest version of pip is available"
 	# $(PIP) install -U pip
-	$(PIP) install --no-index --no-dependencies --find-links file:///$(CACHE)/python -r requirements.txt
+	$(PIP) install --no-index --no-dependencies --find-links file:///"$(CACHE)"/python -r requirements.txt
```
mitechie: I think the quotes here need to go around the /python as well to best work.

So it'd be "file:///$(CACHE)/python"

mitechie:
Thanks so much for the pull request. I think there's one small tweak to get the fix as readable as possible.

aka001:
Thank you for the review. Shall I make a new pull request?

mitechie:
Thanks so much for the update. I'll QA it later today and we'll get this landed.

mitechie:
Mention #279 to make sure it's tied to the issue.

aka001:
Yaa I have mentioned in the title.

mitechie:
This required some more tweaks to work properly. The updated diff is at https://github.com/bookieio/Bookie/commit/7554f8265428dec57e945eb50ef1407694bf51fe but a fresh install in a directory with a space should work correctly now. 

Please look into the notes and other changes in there.

mitechie:
Merged at 7554f8265428dec5

aka001:
Okay got you. Thanks for merging it. :)

PR #288, opened 2014-02-25
Fixed #211, Username case insensitive

### LINKED ISSUE #211: [Auth] Authentication of Username should not Case Sensitive.
For example, as my User ID.

> EdwardChanJW - login successful
> edwardchanjw - login failed

I assumes that the live of bookies are in sync with github issues of Bookie. I only search open issues for duplicate.

sambuddhabasu:
The database migration has been done where the previous usernames are converted to lowercase. Also, the usernames entered are now case insensitive.


bookie/static/js/bookie/view.js
```diff
@@ -9,6 +9,24 @@ YUI.add('bookie-view', function (Y) {
     var _ = Y.substitute,
         ns = Y.namespace('bookie');
 
+    ns.DifferentURLTypes = Y.Base.create('bookie-different-url-type', Y.View, [], {
```
mitechie: The goal is to create a View to handle the JS used in the edit page currently. So the rest of the JS in there will get moved here. From https://github.com/bookieio/Bookie/blob/develop/bookie/templates/bmark/edit.mako#L99

bookie/static/js/bookie/view.js
```diff
@@ -9,6 +9,24 @@ YUI.add('bookie-view', function (Y) {
...
+        /**
+	*
+	* @method _get_different_url
+	* @private
+	*
+	*/
+	_get_different_url: function () {
```
mitechie: what's the 'different' url? I'm not following that naming.

bookie/static/js/bookie/view.js
```diff
@@ -9,6 +9,24 @@ YUI.add('bookie-view', function (Y) {
...
+	    Y.one('input[name="url"]').on(
+	    'onblur',
+	    this._verify_url,
+	    this
+	    );
+	},
+	_verify_url: function (e) {
```
mitechie: so this is going to validate the url. It should read the current value, and then run it against the regex to make sure it's valid or not. Right now it's just getting the url but not doing anything with it.

bookie/static/js/bookie/view.js
```diff
@@ -9,6 +9,24 @@ YUI.add('bookie-view', function (Y) {
...
 
+    ns.DifferentURLTypes = Y.Base.create('bookie-different-url-type', Y.View, [], {
+        /**
+	*
+	* @method _get_different_url
+	* @private
+	*
+	*/
+	_get_different_url: function () {
+	    Y.one('input[name="url"]').on(
```
mitechie: this event can be setup via a Y.View event. See the YUI docs http://yuilibrary.com/yui/docs/view/#handling-dom-events.

There's some other views in the file that demonstrate this as well.

mitechie:
Thanks for this start. I've got a couple of questions and some more links and details to help flesh out the idea farther in the inline comments.

bookie/static/js/bookie/view.js
sambuddhabasu: The function name needs to be changed here. Thanks for pointing that out.

dbversions/versions/44dccb7b8b82_update_username_to_l.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '44dccb7b8b82'
+down_revision = '9f274a38d84'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    connection = op.get_bind()
+    connection.execute("update users set username =
```
mitechie: This is cool and approximately what we need. The big issue is that the query has to work across sqlite, mysql, and postgresql. The best way to write a query like that is to use our DB layer SqlAlchemy to write the query. We can even do it in several steps that way if it helps. Check out http://docs.sqlalchemy.org/en/rel_0_9/core/tutorial.html for notes on the SqlAlchemy expression layer and using it to write queries.

mitechie:
Thanks! This is looking really good and the migration will help a ton. I've added a note on using the SqlAlchemy expression layer to make sure the query is cross db compatible.

mitechie:
landed, thanks so much for going through all of this!

mitechie:
landed.

Issue #273, opened 2014-02-06
Readable page doesn't have title of content


craigmaloney:
Example: https://bmark.us/bmark/readable/3467f263dbe385

Would be nice to have the title of the page so when I print it out (save as PDF, actually) it would use the title for the filename.

mitechie:
How does it not have the title? It's there on the page?

mitechie:
or do you mean the html page title so that the browser prints it in the margin up top?

craigmaloney:
Yeah, it's the title at the top of the HTML page. Was hoping it would
more closely match the title of the content of the page, not the site
itself.

mitechie:
Landed and pushed live.

PR #289, opened 2014-02-25
Fix for Issue #273

badanomaly:
Now the title of the page shows bmark description.
If no description is available then it shows the bmark url.


mitechie:
Code looks good. I'll defer to Craig that it solves his problem to close the bug.

craigmaloney:
Looks good to me. Thanks so much!

mitechie:
Merged as 44b076eedc1

Thanks for the great work! Hopefully it was useful to learn a few things.

Issue #291, opened 2014-02-25
Url in the edit page not properly aligned 


saurabh95:

saurabh95:
I am currently working on it

mitechie:
This is landed. Closing

PR #294, opened 2014-02-25
Fixed #292 #291

### LINKED ISSUE #292: The Url in the edit page is simply text
The Url in the edit page is simply text which can be converted to a working URL so that  user can directly click on that to go to the URL instead of copying it

saurabh95:
Added the href tag for the URL in the edit page and also aligned it properly


mitechie:
Thanks, this looks good! Please update the CREDITS file and repush your branch so that the pull request can add it. I can then merge it.

saurabh95:
I have added my name in CREDITS.txt :)

mitechie:
Landed, thanks for the update!

PR #301, opened 2014-02-26
Issue Resolved #296

### LINKED ISSUE #296: Add new Bookmark page throws error when a previously existing URL is added.

rishirajsinghjhelumi:


mitechie:
This looks like a really solid change. Are you up to trying to write a test for it? We should add a test that verifies that you get the message when posting a url that already exists. This would go with the other bmark tests in https://github.com/bookieio/Bookie/blob/develop/bookie/tests/test_webviews/test_webviews.py#L86

rishirajsinghjhelumi:
I will surely write tests for it. I will go through the above files and get back to you in case I don't understand anything.

bookie/tests/test_webviews/test_webviews.py
```diff
@@ -108,6 +108,38 @@ def test_manual_entry_error(self):
             })
         self.assertIn('not valid', res.body)
 
+    def test_existing_url_entry_error(self):
+        """ Verify whether the User has recieved the message that a previous URL already exists """
+        self._login_admin()
+
+        _test_url = u"http://bmark.us/test"
```
mitechie: there's no need to underscore these two lines.

bookie/tests/test_webviews/test_webviews.py
```diff
@@ -108,6 +108,38 @@ def test_manual_entry_error(self):
...
+        _existing_url_message = "URL already Exists"
+
+        #Add The Bookmark Once
```
mitechie: always have a space after the # in the comment and make it a real sentence. 

# Add the bookmark once.

bookie/tests/test_webviews/test_webviews.py
```diff
@@ -108,6 +108,38 @@ def test_manual_entry_error(self):
...
+                'description': '',
+                'extended': '',
+                'tags': ''
+            })
+        self.assertEqual(
+            res.status,
+            "302 Found",
+            msg='recent status is 302 Found, ' + res.status)
+
+        #Add the Bookmark Again
```
mitechie: # Add the bookmark again.

bookie/views/bmarks.py
```diff
@@ -165,6 +165,17 @@ def edit_error(request):
     with ReqAuthorize(request, username=rdict['username']):
         if 'new' in request.url:
             try:
+                try:
+                    bmark = BmarkMgr.get_by_url(post['url'])
+                except:
+                    bmark = None
+                if bmark:
```
mitechie, line 172: can we tweak the spacing here? It should be four spaces indent and it looks off here. All indentation should be spaces. Maybe there's some tabs in here?

mitechie:
Landed, thanks for the changes!

Issue #298, opened 2014-02-26
empty Name and email does not show error while editing details in the account tab


saurabh95:
Empty name and email does not show any error and also it does not check whether it is a valid email or not

saurabh95:
working on it

mitechie:
Landed this update.

PR #303, opened 2014-02-26
Fixed issue #298

saurabh95:
Now the email is validated and also user is informed if he has entered empty name or email


mitechie:
Thanks for working hard on this. I've landed another branch that's very similiar as to the type of setup and testing required. Please take a look at how it sets up the extra html in the test file and then fakes out the AJAX call with data and runs the tests after that completes. 

https://github.com/bookieio/Bookie/commit/6195cab1190598d1c40985feab0a249d1242769c

PR #304, opened 2014-02-26
Fixes #290

### LINKED ISSUE #290: timezone bug in the tests
http://paste.ubuntu.com/6995880/

pqrth:
Since the test is system dependent, absolute date values replaced with
relative one in chrome_data_test


mitechie:
Thanks so much for the test fix!

PR #306, opened 2014-02-27
Fixed issue#254

badanomaly:
Fix for issue#254. Now search returns a "good" error message when accessing an out of bound page through the api.


bookie/views/api.py
```diff
@@ -513,11 +513,16 @@ def search_results(request):
...
-    res_list = searcher.search(phrase,
+    try:
+        res_list = searcher.search(phrase,
                                content=search_content,
                                username=username if with_user else None,
                                ct=count,
                                page=page)
+    except ValueError, exc:
+        request.response.status_int = 404
+        ret = {'error': "Bad Request:Page number out of bound"}
```
mitechie: let's put a space after the colon here.

mitechie:
This looks like a great fix. Are you up for writing a test for it? 

The link below is to an existing search test and in your case you could tweak something like that and add the offset "page=10" or what not to the request to generate the error. Then you'd be verifying that you get the 404 and the json error message returned. 

https://github.com/bookieio/Bookie/blob/develop/bookie/tests/test_api/test_base_api.py#L369

badanomaly:
Closing this pull request because i again messed up my working directory and there are a lot of unnecessary commits associated with this one.

PR #309, opened 2014-02-27
fix repository url in git clone instructions in README.rst

dyladan:
It currently points to mitechie when it should point to bookieio


mitechie:
Thanks!

Issue #311, opened 2014-02-28
502 bad gateway on submitting import form without a file to upload


dyladan:
I think it is related to this file but since i have very little python and even less pyramid i'm not sure

https://github.com/bookieio/Bookie/blob/develop/bookie/views/utils.py#L58

sambuddhabasu:
I am currently working on this bug.

preddy5:
@sammyshj hey im already working on it

sambuddhabasu:
@pradyu1993 Sorry, no one had commented on this bug, so I had started working on it. Already done with the patch. Testing that nothing breaks. Will be done with the pull request in a while.
I guess it would be good if you start working on some other issue :)

mitechie:
Landed

PR #314, opened 2014-02-28
issue #311

preddy5:


bookie/views/utils.py
```diff
@@ -57,7 +57,7 @@ def import_bmarks(self):
                 # we have some posted values
                 files = post.get('import_file', None)
 
-                if files is not None:
+                if files:
```
preddy5: check for unicode string, "is not None" works for non-unicode string

bookie/views/utils.py
```diff
@@ -77,7 +77,7 @@ def import_bmarks(self):
                         location=self.request.route_url('user_import',
                                                         username=username))
                 else:
-                    msg = self.request.session.pop_flash()
+                    msg = "Please Add a File" ##self.request.session.pop_flash()
```
preddy5: @mitechie we need to create a Configurator instance to use session I wanted to take your view before creating it, but for now the bug is solved with the dummy message, once we decide on the configurator instance we could solve the bug completely

mitechie:
Thanks, landed this. Appreciate the quick fix.

PR #319, opened 2014-03-01
Added the check for valid email and empty name and the required test for Account information change

saurabh95:


bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
             e.preventDefault();
             Y.one('#account_msg').hide();
 
+            var name = Y.one('#name').get('value');
```
mitechie, line 1449: looks like a space is off here?

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
...
+	    var email = Y.one('#email').get('value');
+	    var re = RegExp(['/^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)'
```
mitechie: ah, much nicer thanks.

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
...
+		            ,'|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])'
+		            ,'|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/'].join(''));
+	    // validate email and username
+	    if(!name || !email || !re.test(email)) {
+		if(!name) {
+                    that._show_message('Please enter valid Name', true);
+		}
```
mitechie: the else if should be on the same line as the closing }

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
             e.preventDefault();
             Y.one('#account_msg').hide();
 
+            var name = Y.one('#name').get('value');
+	    var email = Y.one('#email').get('value');
+	    var re = RegExp(['/^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)'
+		            ,'|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])'
+		            ,'|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/'].join(''));
+	    // validate email and username
+	    if(!name || !email || !re.test(email)) {
```
mitechie: all of the if and else if should have a space between the if and the opening paren (

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
...
+                    that._show_message('Please enter valid Email', true);
+                    Y.one('#email').set('value','');
+		}
+	    }
             // add the password data to the cfg passed to the api
-            api_cfg = Y.merge(api_cfg, {
-                name: Y.one('#name').get('value'),
-                email: Y.one('#email').get('value')
-            });
+	    else {
```
mitechie: like the other, it should be } else {

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,41 @@ YUI.add('bookie-view', function (Y) {
...
-            });
+                api = new Y.bookie.Api.route.UserAccountChange(api_cfg);
+                api.call({
+                    success: function (data, request) {
+                        that._show_message('Account updated...', true);
+                    },
+                    error: function (data, status_str, response, args) {
+                        console.log(data);
+                        console.log(response);
+                        that._show_message('Some error ocurred while \
```
mitechie: if you break this after the ( and newline it might fit the text on one line without the . Can you try that?
    that._show_message(
        'Some error occurred...'

mitechie:
Thanks! This looks a lot better. I'll pull it down to test and land later today.

bookie/static/js/bookie/view.js
```diff
@@ -1446,22 +1446,42 @@ YUI.add('bookie-view', function (Y) {
...
+                api = new Y.bookie.Api.route.UserAccountChange(api_cfg);
+                api.call({
+                    success: function (data, request) {
+                        that._show_message('Account updated...', true);
+                    },
+                    error: function (data, status_str, response, args) {
+                        console.log(data);
+                        console.log(response);
+                        that._show_message(
+                            'Some error ocurred while making changes...', true
```
mitechie: should the second param be false then since this is an error? This should fix the test then where you can verify it is indeed a test.

mitechie:
Landed this change without the email regex validation. I updated the tests to work and make sure both an empty name and empty email triggered the hint.

The regex did not permit email addresses with a + in them. This is a standard way for gmail (and google apps users) to identify where an email comes from and is a valid character for an email. There were also no tests verifying the regex worked at all. We can look to add this back in as a follow up branch.

mitechie:
This is now running in production on bmark.us

Issue #87, opened 2011-06-20
start running daily cron jobs to store counts per user


mitechie:
We want to be able to provide some trend graph info to users. We want to run daily stats on how many total bookmarks and tags they have in the system.

Then we can add a stats page on their account pages that can track that over time. Might be interesting, especially once we start allowing trimming of bad links/etc. Might stick that in there as well. Count the number of readable that ended in 404, unresolved, etc.

mitechie:
I basically want to get celery and rabbitmq going. These tasks will be fired off via scheduled jobs there. 

So the idea would be to create a 24hr task that inits a task per user to gather their stats into the stats log.

mitechie:
This should be run from the new celery system. Basically a scheduled item to calc the total for each user that then fires of subtask items for each user in the system to go do the actual counting.

sambuddhabasu:
I am working on this issue.

mitechie:
This is landed and bookmark counts should start today.

PR #320, opened 2014-03-01
Bookmark count for individual user with tests

sambuddhabasu:


sambuddhabasu:
Fixed issue #87

mitechie:
I've landed this with some updates.
- I've used the existing UserMgr.get_list(active=True) to find the users. Finding a list of users should be the UserMgr's job, not Bmarks.
- I've tweaked the stats method to not do the query. The stats should know as little as possible and only deal with the stats database. The username is passed in now from the task and it only saves a stats record and does not touch other areas of the db.
- Updated the test to check that the numbers for all three sample users are correct.

mitechie:
This is now running in production on bmark.us

sambuddhabasu:
Thanks and thats some great news!

PR #321, opened 2014-03-02
User bookmark count api

sambuddhabasu:


bookie/models/__init__.py
```diff
@@ -6,6 +6,7 @@
 
 from datetime import datetime
 from datetime import timedelta
+from calendar import monthrange
```
mitechie: these should be sorted alphabetically. calendar comes before datetime.

bookie/models/__init__.py
```diff
@@ -469,6 +470,26 @@ def user_dump(username):
             filter(Bmark.username == username).all()
 
     @staticmethod
+    def user_count_bmark(username, start_date=None, end_date=None):
+        """Get a list of user bookmark count
+        """
+        from bookie.models.stats import StatBookmark
+        days = 30
```
mitechie: let's make this a module level constant? So up to the top something like STATS_WINDOW = 30

bookie/models/__init__.py
```diff
@@ -469,6 +470,26 @@ def user_dump(username):
...
+        qry = StatBookmark.query.\
```
mitechie: this shouldn't talk to the db. It should talk to the StatsMgr. We might have to add a method on there to do the query for us.

bookie/models/__init__.py
```diff
@@ -469,6 +470,26 @@ def user_dump(username):
...
+            filter(StatBookmark.attrib == username)
+        if not start_date:
```
mitechie: this gets a little dense to read and follow. Maybe some comments along the way to help out?

bookie/models/__init__.py
```diff
@@ -469,6 +470,26 @@ def user_dump(username):
             filter(Bmark.username == username).all()
 
     @staticmethod
+    def user_count_bmark(username, start_date=None, end_date=None):
```
mitechie: this should only be in the stats model.

bookie/models/__init__.py
sambuddhabasu: changes have been done in https://github.com/sammyshj/Bookie/commit/e0edb6a835d7fe829274f873ffb562390ea18b34

bookie/views/api.py
```diff
@@ -365,6 +365,33 @@ def bmark_recent(request, with_content=False):
     })
 
 
+@view_config(route_name="api_count_bmarks_user", renderer="jsonp")
+@api_auth('api_key', UserMgr.get, anon=False)
+def user_bmark_count(request):
+    """Get a list of bookmark count for the api call"""
```
mitechie: Get the user's daily bookmark total for the given time window. 

Might be a better description?

bookie/views/api.py
```diff
@@ -365,6 +365,33 @@ def bmark_recent(request, with_content=False):
...
+    rdict = request.matchdict
+    params = request.params
+
+    username = 'user_bookmarks_' + request.user.username
```
mitechie: so this should build the string using the constant in the models/stats.py like we do in there already. That way, if we tweak it, there's only one place to make the change.

bookie/views/api.py
```diff
@@ -365,6 +365,33 @@ def bmark_recent(request, with_content=False):
...
+    start_date = params.get('start_date', None)
+    end_date = params.get('end_date', None)
+    bmark_count_list = BmarkMgr.user_count_bmark(
```
mitechie: so then this will turn into the StatsMgr

bookie/views/api.py
```diff
@@ -365,6 +365,33 @@ def bmark_recent(request, with_content=False):
...
+    )
+
+    result_set = []
+    for res in bmark_count_list:
+        return_obj = dict(res)
+
+        result_set.append(return_obj)
+
+    return _api_response(request, {
+        'bookmark_count': result_set
```
mitechie: hmm, the api method is bmarkcount. I'd just make this count and not bookmark_count.

bookie/models/stats.py
```diff
@@ -58,6 +61,17 @@ def get_stat(start, end, *stats):
         return qry.all()
 
     @staticmethod
+    def get_user_stat(username, start_date=None, end_date=None):
+        """Fetch the bookmark count for the user from the stats table"""
+        qry = StatBookmark.query.\
+            filter(StatBookmark.attrib==username)
```
mitechie: The attrib is stored as  USER_CT.format(username) so when we query the db to get it out, we need to build that matching string back up again. Right now it's just querying for the uesrname without the rest of the prefix.

bookie/models/stats.py
```diff
@@ -58,6 +61,15 @@ def get_stat(start, end, *stats):
         return qry.all()
 
     @staticmethod
+    def get_user_bmark_count(username, start_date, end_date):
+        """Fetch the bookmark count for the user from the stats table"""
+        qry = StatBookmark.query.\
+            filter(StatBookmark.attrib==USER_CT.format(username)).\
+            filter(StatBookmark.tstamp>=start_date).\
+            filter(StatBookmark.tstamp<=end_date)
+        return qry.all()
```
mitechie, line 73: we should probably add an order_by on here so that the results always come out in order.

mitechie, line 73: http://docs.sqlalchemy.org/en/rel_0_9/orm/tutorial.html#querying for some notes on querying

sambuddhabasu, line 73: sure an order_by tstamp would be a good idea

bookie/models/stats.py
```diff
@@ -95,6 +107,24 @@ def count_user_bookmarks(username):
...
+            if not end_date:
+                """If both start_date and end_date are None,
+                end_date will be the current date"""
+                end_date = datetime.utcnow()
+            start_date = end_date - timedelta(days=STATS_WINDOW)
+        elif start_date and not end_date:
+            if start_date.day == 1:
+                """If the starting day is 1, stats of the month is returned"""
+                global STATS_WINDOW
+                STATS_WINDOW = monthrange(start_date.year, start_date.day)[1] - 1
```
mitechie: Oh, I see. Because you're updating the value here.

mitechie: I don't think you should update this value. You can use it or use another variable. If someone makes an api call and changes it to 60 days, the next api call will have STATS_WINDOW = the 60 days. It'll break everyone else's api call.

sambuddhabasu: Changes were made. Thanks for pointing that out

mitechie:
superceded

Issue #274, opened 2014-02-08
Users should have a delete button for their bookmarks


mitechie:
When a user mistakenly imports something they did not intend to be public or decides they don't want their data in the bookie install any more we should let them delete their account/data.

carpedm20:
I think there is already delete features in account/recent page. Do you mean a page which will be shown right after a user import a .html file and display only imported bookmarks to select which bookmarks will be deleted or not? Or do you mean another features?

mitechie:
This is meant to be a global delete. From a user's account page should be a "delete all of my bookmarks" button so that they can remove all their bookmarks from the system in one swoop.

I'd imagine that this would then fire off a background job in bcelery/tasks to do the actual work of removing the users bookmarks from the system in batches.

preddy5:
I am currently working on this bug

kshitijiiit:
Solved.

mitechie:
Fix released.

PR #325, opened 2014-03-04
Added functionality for deleting all bookmarks for a user

kshitijiiit:
Fixed Issue #274. (Delete all bookmarks functionality)

Steps:
Go to Account -> Delete All Bookmarks.


bookie/models/__init__.py
```diff
@@ -567,6 +567,16 @@ def count(username=None, distinct=False, distinct_users=False):
             qry = DBSession.query(Bmark.username).distinct()
         return qry.count()
 
+    @staticmethod
+    def delete_all_bookmarks(username):
```
mitechie, line 571: why do you need a method on the model to do the delete? I would expect this to have the query to do the delete and the task to call the model here.

bookie/views/bmarks.py
```diff
@@ -253,3 +253,18 @@ def readable(request):
...
             return HTTPNotFound()
+
+@view_config(route_name="user_delete_all_bookmarks", renderer="/accounts/index.mako")
+def delete_all_bookmarks(request):
+    """Delete all bookmarks of the current user"""
+    rdict = request.matchdict
+    with ReqAuthorize(request, username=rdict['username'].lower()):
+	username = request.user.username
+        if username:
+	    BmarkMgr.delete_all_bookmarks(username)
```
mitechie: I think it would be better if this loaded the task and called the task, then the task would call the model which would do the db query to remove the bookmarks.

kshitijiiit: Okay, I see your point. I'll make the changes right away and send another pull request.

bookie/models/__init__.py
kshitijiiit, line 571: Okay. Also, should the task be called with the delay() option?

mitechie:
This is a really solid start. I think to do this task well we should have a confirmation dialog in JS with an "are you sure" or maybe just that when they hit the button, the color/text changes on it to be a "Confirm" so that the user must double act to delete their bookmarks. 

I had thought about making the delete all bookmarks an api call, but now that I think it through, it's not something you'd do repeatedly. Perhaps that doesn't make much sense. 

When the user does confirm, we should notify them that the delete request has been queued and will be acted upon shortly. Because of the way the task system works, their bookmarks might not delete until minutes or hours later. We could have the form POST back to the account page with a url query added on? https://bmark.us/admin/account?delete_all=true

If that's set then instead of the "delete all" form they get the notice "A request to delete all bookmarks is pending in the queue"? What are your thoughts?

Finally, we need to add tests to the system for these. There should be a test for the model in the test_models that makes sure all bookmarks for a user are removed as well as for the view that verifies that it loads the correct form. Let me know if you need some help getting into the tests as I know sometimes it's challenging.

kshitijiiit:
Yes, I too was thinking of a confirmation to delete all the bookmarks. I think the idea on POSTing back to the account page with the message seems reasonable. Should we notify the user once the task has completed too?
About the tests, I really haven't looked at it yet. I plan to write some tests for it next. I'll ping you on IRC in case I get stuck somewhere. :)

mitechie:
We could notify when it's done, but I think seeing all their bookmarks gone will be proof. We can always tweak it and make things more complicated as a follow up. For a first pass I think the POST and message that we've accepted the task is a great first step.

bookie/models/__init__.py
mitechie, line 571: Yes, the .delay() means it won't wait for the task to happen. It'll just return and schedule it.

preddy5:
@mitechie I think rather than a simple Yes or No i think a form asking for username, like in github to delete a repo we need to type its name to delete, cause this operation will delete all the bookmarks that'll be a good idea. thoughts?

bookie/models/__init__.py
mitechie, line 571: http://celery.readthedocs.org/en/latest/userguide/calling.html for some docs on calling tasks with celery. 

http://celery.readthedocs.org/en/latest/getting-started/first-steps-with-celery.html#calling-the-task for the complete 101 notes.

kshitijiiit, line 571: Oh, cool. This is really helpful!

mitechie:
Hmm, a confirmation field to fill in isn't a bad solution and removes the need for a JS driven confirmation for now. 

Maybe something like "To remove all of your bookmarks in the system enter the word "Delete" in the field below and hit submit."

_________ [submit]

What do you think +kshitijiiit?

kshitijiiit:
Yeah, that's a nice way to avoid JS. Also, it prevents user from accidentally deleting all bookmarks by clicking twice.

preddy5:
@mitechie Ya exactly something like that, implementing it JS will also be cool

preddy5:
@mitechie but then probably we'll need to use js libraries that not being used right now?

mitechie:
Not at all, we have all of YUI available for JS libraries. It does everything we'd need. We'd just add the functionality to the Account View in the bookie/static/js/view.js file. 

I think we can implement it without any JS though, just verify that the form is posted back to the account page with the right value and we can reload it and show the notification to the user. 

If we call the delete all from JS then we should move the functionality to the API and call it from JS like the other API calls we have currently.

preddy5:
got it, Thanks

kshitijiiit:
Updated the code structure and added relevant tests. All commits are squashed into one. Kindly review.

bookie/routes.py
```diff
@@ -49,6 +49,7 @@ def build_routes(config):
                      "{username}/edit_error/{hash_id}")
     config.add_route("user_bmark_new", "{username}/new")
     config.add_route("user_bmark_new_error", "{username}/new_error")
+    config.add_route("user_delete_all_bookmarks", "{username}/account/delete_all_bookmarks")
```
mitechie, line 52: Hmm, do we still need this route? Did we end up posting back to the account page?

bookie/templates/accounts/index.mako
```diff
@@ -112,9 +122,17 @@ ${password_reset(user, reset=False)}
         <textarea style="display: none; opacity: 0; width: 25em; height: 8em; padding: 1em; margin-top: 1em;" id="bookmarklet_text">javascript:(function() {location.href='${request.host_url}/${request.user.username}/new?url='+encodeURIComponent(location.href)+'&description='+encodeURIComponent(document.title)}())</textarea>
     </div>
     </p>
+</div>
 
-
+<div class="form">
+    <div class="heading">Delete all bookmarks</div>
+     <form method="post" action="${request.route_url('user_delete_all_bookmarks', username=request.user.username)}">
```
mitechie, line 129: ah ok, you post to this route and then end up back here?

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -10,7 +12,6 @@
 from bookie.tests import gen_random_word
 from bookie.tests import TestDBBase
 
-
```
mitechie, line 13: this should fail make lint. You need two blank lines between imports and the first function/class.

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -117,3 +118,31 @@ def test_per_user(self):
...
         ct = BmarkMgr.count(username=usercommon.username)
         self.assertEqual(2, ct, 'We should have a total of 2: ' + str(ct))
+
+    def test_delete_all_bookmarks(self):
+        """Testing working of delete all bookmarks
+                Case 1: No bookmark present
+                Case 2: One bookmark present
+                Case 3: Multiple bookmarks present"""
+
+        l = [0, 1]
```
mitechie: please don't use single variable names. 'l'? It looks like you're creating a user for every number in l?

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -117,3 +118,31 @@ def test_per_user(self):
...
+        for i in range(10):
+            l.append(randint(10, 100))
```
mitechie: so we have users 0, 1, and 10 random ints between 10 and 100?

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -117,3 +118,31 @@ def test_per_user(self):
...
+
+        for i in range(len(l)):
```
mitechie: ok, so that number is actually the number of bookmarks they have. So I'd rename l bmark_counts and then for count in bmark_counts

bookie/tests/test_webviews/test_account.py
```diff
@@ -0,0 +1,75 @@
...
+from bookie.tests import TestViewBase
+from bookie.tests import gen_random_word
+
+LOG = logging.getLogger(__name__)
+
+class AccountViewsTest(TestViewBase):
+    """Test the account web views for a user"""
+
+    def test_delete_all_bookmarks_with_correct_confirmation(self):
+        """
```
mitechie: Check out pep357 that governs proper docstrings. They should always start on the line with the quotes.

http://legacy.python.org/dev/peps/pep-0257/

bookie/tests/test_webviews/test_account.py
```diff
@@ -0,0 +1,75 @@
...
+                'username': 'admin',
+                'delete': gen_random_word(10),
+            })
+
+        self.assertEqual(
+            res.status,
+            "200 OK",
+            msg='recent status is 200, ' + res.status)
+        self.assertTrue(
+            "Delete request NOT initiated due to incorrect confirmation." in res.body,
```
mitechie: How about:

"Delete request not confirmed. Please make sure to enter 'Delete' to confirm."

bookie/views/bmarks.py
```diff
@@ -253,3 +253,27 @@ def readable(request):
             }
         else:
             return HTTPNotFound()
+
+@view_config(route_name="user_delete_all_bookmarks", renderer="/accounts/index.mako")
+def delete_all_bookmarks(request):
+    """Delete all bookmarks of the current user"""
+    rdict = request.matchdict
+    post = request.POST
+    with ReqAuthorize(request, username=rdict['username'].lower()):
```
mitechie, line 264: nice on catching the .lower

bookie/views/bmarks.py
```diff
@@ -253,3 +253,27 @@ def readable(request):
...
+	username = request.user.username
```
mitechie: Do we need this? If there's no username then the request won't authorize and you'll get a 403?

mitechie:
This looks great. Just a few small nitpicks. Trying out the changes locally.

mitechie:
Tested this out and it works great. A little code gardening and I'll land this. Thanks so much for the hard work on this!

kshitijiiit:
Pleasure is mine. It was a good learning experience for me :)

bookie/templates/accounts/index.mako
kshitijiiit, line 129: Yes.

bookie/tests/test_models/test_bmarkmgr.py
kshitijiiit, line 13: Right. I'll run lint and make the changes.

bookie/tests/test_models/test_bmarkmgr.py
kshitijiiit: Yeah, will update this.

bookie/tests/test_models/test_bmarkmgr.py
kshitijiiit: Ok, I will change the variable name. Yes, for every element in the list I am repopulating the database with bookmarks.

mitechie:
I've landed this in develop. 

I did tweak it to make sure that the tags are removed as well and altered the test a bit to be able to demonstrate that it does not delete all bookmarks in the system, but only those for each user. 

Thanks so much for this great branch!

Issue #329, opened 2014-03-05
stats/bmarkcount api endpoint fails when specifying the start_date param


widox:
I tried hitting this API endpoint:
`https://bmark.us/api/v1/<user>/stats/bmarkcount?api_key=<key>&start_date=2014-01-01`

and got a `502 Bad Gateway` response, however the docs say that the `start_date` and `end_date` params are just plain dates.

mitechie:
Thanks for this Matt. I'll talk to sammyshj about it.

sambuddhabasu:
Working on it right now.

mitechie:
fix released

PR #330, opened 2014-03-05
Fixed issue #329. User bookmark count works with only date now

sambuddhabasu:


bookie/models/stats.py
```diff
@@ -106,9 +106,11 @@ def count_importer_depth():
     def count_user_bookmarks(username):
         """Count the total number of bookmarks for the user in the system"""
         total = BmarkMgr.count(username)
+        tstamp = datetime.utcnow()
         stat = StatBookmark(
             attrib=USER_CT.format(username),
-            data=total
+            data=total,
+            tstamp=datetime(tstamp.year, tstamp.month, tstamp.day)
```
mitechie: you don't have to do this. When you query the datetime you'll query dates >= day and it'll use 00:00:00 for  the time automatically.

bookie/models/stats.py
```diff
@@ -116,14 +118,17 @@ def count_user_bookmarks(username):
     def count_user_bmarks(username, start_date=None, end_date=None):
         """Get a list of user bookmark count"""
         if start_date:
-            start_date = datetime.strptime(start_date, '%Y-%m-%d %H:%M:%S')
+            start_date = start_date.split(' ')
```
mitechie, line 119: why the split?

bookie/tests/factory.py
```diff
@@ -85,6 +85,7 @@ def make_user_bookmark_count(username, data, tstamp=None):
     """Generate a fake user bookmark count for testing use"""
     if tstamp is None:
         tstamp = datetime.utcnow()
+        tstamp = datetime(tstamp.year, tstamp.month, tstamp.day)
```
mitechie: I don't think this change is required.

bookie/models/stats.py
sambuddhabasu, line 119: The split was used because when the date is passed from other functions, start_date is taken as a string like '2014-04-03 00:00:00'. The split separates the date. If the split is not specified, an error is encountered saying 'unconverted data remains'

sambuddhabasu:
@mitechie The necessary changes were done. I guess the pull request can be merged now?

bookie/tests/test_api/test_base_api.py
```diff
@@ -526,7 +526,7 @@ def test_start_end_defined(self):
         """Test getting a user's bookmark count over a period of time when
         start_date is None and end_date is defined"""
         test_dates = self._setup_user_bookmark_count()
-        test_end_date = datetime(2013, 12, 29, 8, 8, 8)
+        test_end_date = datetime(2013, 12, 29)
         res = self.testapp.get(u'/api/v1/admin/stats/bmarkcount',
                                params={u'api_key': API_KEY,
                                        u'end_date': test_end_date},
```
mitechie: I think the tests should be updated to use the string you would expect a user to use. So u'end_date': u'2013-12-29'

sambuddhabasu: Yeah this would make the tests more like the real world situations. Thanks, will be changing them now

sambuddhabasu: Also, in the assertEqual(), we would have something like  data['start_date'], u'2013-11-29 00:00:00'

sambuddhabasu:
Fixed issue #329

mitechie:
I updated this for one small issue. The end_date ends up being today but at 00:00:00 so it won't grab stats from today at 3am, 6pm, etc. I increment the end day by one day so that it's tomorrow at midnight which works for the behavior of getting states from today backwards.

Thanks for this update!

PR #332, opened 2014-03-05
Issue#331

badanomaly:
Fixed the issue of bookmarks without tags not being present in the export.
Added the condition to show links for bookmarks in place of description, if a bookmark doesn't have a description associated with it, in the export file.
Added tests.


mitechie:
Thanks for this! I've updated the query and tests to work out properly. Please take a look at the commit that landed to see what was updated.

Landed in trunk and will be deploying shortly.

PR #333, opened 2014-03-05
added code to verify username

rakeshmahadasa:


bookie/views/auth.py
```diff
@@ -193,8 +193,13 @@ def reset(request):
         password = params.get('new_password', None)
         new_username = params.get('new_username', None)
         error = None
-
-        if not UserMgr.acceptable_password(password):
+        
+        #Check whether username already exists.
```
mitechie: Make sure to have a space after the # to start a comment.

bookie/views/auth.py
```diff
@@ -193,8 +193,13 @@ def reset(request):
...
+        if UserMgr.get(username) == None:
```
mitechie: If you compare against None, then make sure to use 'is' vs ==

```
if UserMgr.get(username) is None:
```

or you I think you could just do

```
if UserMgr.get(username)
```

mitechie:
This is very cool. A couple of small nitpicks on the syntax. 

Any chance you're up for writing a unit test to verify that this works correctly? I'd be happy to help guide you through it.

rakeshmahadasa:
Thank you Rick for your guidance. I would love to work on writing unit test. give me a brief idea.

mitechie:
Been superceded

PR #335, opened 2014-03-07
Username verification needed 

rakeshmahadasa:


mitechie:
Updated and landed this change. I fixed up the test so please make sure to check it out in commit: https://github.com/bookieio/Bookie/commit/de666576d0d9cac4c871b1d80ee707d3b2d83903

We had to adjust how the user was created, the url hit in the test, etc.

PR #337, opened 2014-03-07
fix url alignment on edit page

dyladan:
On the page to edit bookmarks the url field alignment is wonky. This makes it line up with it's label.


dyladan:
This fixes #291

mitechie:
Thanks for the branch. Landed in trunk.

PR #338, opened 2014-03-07
fixed empty suggested-tags in edit

dyladan:
On the edit page, if there are no suggested tags, it still displays "suggested tags:" and an empty `<a>` tag that directs back to the same edit page. this fix makes the label and the `<a>` tag hide if there are no suggested tags


dyladan:
Just added a commit so that if a bookmark already has a tag it will not be suggested again

dyladan:
uncommented my code because i'm stupid and pushed with it commented out.
squashed into a single commit

mitechie:
Mind adding a test for this? There's some tests for the add bookmark work in https://github.com/bookieio/Bookie/blob/develop/bookie/tests/test_webviews/test_webviews.py#L86

dyladan:
I'll do this tonight

mitechie:
Cool, make sure to check out the @patch use in bookie/tests/test_auth/test_reset.py to help with mocking out the suggestion calls in tests.

mitechie:
Didn't actually need the check. tag_suggest != []: resolves to if tag_suggest as well. The duplication work is valid and I updated an existing test to verify that was working and landed this change.

PR #339, opened 2014-03-07
Made testcases pytest compatible

### LINKED ISSUE #323: port to using py.test for the python test suite `make test`
What the title says. Everyone says it's the thing to use.

sanchitgangwar:
Fixes #323. Had to make some `from bookie.bcelery import tasks` statements local to method definitions. The reason, as I found out, was that importing `tasks` lead to execution of some global statements which prevented pytest from working.


mitechie:
Thanks for this! I've landed the branch along with some lint changes and such. I ended up moving the bcelery imports back to the outer file. I didn't get any issues. If you could provide some details on why those needed to be moved I can see if I can reproduce, but for now it's working as is.

sanchitgangwar:
Yes, it is working, strangely. No worries, then.

Issue #196, opened 2012-11-11
user based click count isn't processed


mitechie:
I a url is clicked on from a user's home page, it should increment the global count on that bookmark as well as the user one.

so /admin/redirect/hashid

should increment two different counters. 

This currently fails, perhaps because the routes should be in reverse order, going from most specific to least. If we always hit the first route it never populates the {username} in the matchdict.

kshitijiiit:
I am currently working on this.

kshitijiiit:
Solved!

PR #341, opened 2014-03-08
fixed : bookmark count for users incremented

kshitijiiit:
Issue #196 fixed.


mitechie:
Doh! Very cool, thanks for the fix!

PR #342, opened 2014-03-08
JS Tests for user stats

sambuddhabasu:


bookie/static/js/tests/test_user_stats.js
```diff
@@ -0,0 +1,57 @@
...
+                url: 'http://127.0.0.1:6543/api/v1',
+                username: 'admin',
+                api_key: '123456'
+	    };
+            var user_stats = Y.bookie.Api.route.UserBmarkCount(API_CFG);
+            var user_stats = new Y.bookie.UserBmarkCountView({
+                api_cfg: API_CFG,
+                container: Y.one('body')
+            });
+            user_stats.render();
```
mitechie: I don't see any test assertions in this test? Is there something to the test I'm missing? So after render I'd think there's be assertions that there were HTML nodes in the body that the charting library creates?

sambuddhabasu: the current render function calls the UserBmarkCount._fetch_dataset() which results in an error during the api.call({})

bookie/static/js/tests/test_user_stats.js
```diff
@@ -0,0 +1,56 @@
...
+
+        testApiExists: function() {
+            Y.Assert.isObject(Y.bookie.Api,
+                              "Should find an object for Api module");
+        },
+
+        test_error_message: function() {
+            var userstats_msg = Y.one('#userstats_msg');
+            var API_CFG = {
+                url: 'http://127.0.0.1:6543/api/v1',
```
mitechie: so we don't want to make real api calls to real urls here. This is part of the mocking setup we talked about. 

See https://github.com/bookieio/Bookie/blob/develop/bookie/static/js/tests/test_view.js#L298

Notice that it creates a fake response and calls the 'error' callback. In this way then you should not make a real http request out to the service.

bookie/static/js/tests/test_user_stats.js
mitechie: so we should have two test functions. One for the good case and one for the error case like you've got. Each should have two different mocked functions. One calls the success and one for the error case.

sambuddhabasu:
The JS tests have been updated to now send fake error and success responses. The tests run perfectly on a local instance now.

mitechie:
Rebased and landed. Thanks for the tests!

Issue #249, opened 2013-10-10
use chrome data sync to track bookmarked urls for user notice


mitechie:
http://developer.chrome.com/extensions/storage.html#property-sync

mitechie:
SYNC YOUR OPTIONS SETTINGS!!!!!!!!

PR #345, opened 2014-03-10
Chrome Storage implementation in place of local storage and option to enable/disable sync

suryatech:
In order to sync the user credentials like (api_key, api_username, api_url, cache_content) we need to use the chrome.storage.sync feature instead of local storage.

Also, with chrome.storage.local/sync boolean variables can be stored as is without converting them to strings and thus eliminating the need to convert them back to booleans. 

Following commit implements the same and fixes some typos in the documentation along with some fresh documentation. 

Addresses modified #249 as per discussion.

### DIFF
--- a/bookie/static/js/bookie/model.js
+++ b/bookie/static/js/bookie/model.js
@@ -763,7 +763,6 @@ YUI.add('bookie-model', function (Y) {
                 }
                 that.set(key, ret);
 
-<<<<<<< HEAD
                 if(callback)
                     callback();
             }
@@ -779,7 +778,7 @@ YUI.add('bookie-model', function (Y) {
                     });
                 }
             }
-=======
+
             function onSyncRead(){
                 if (sync && (key === 'api_key' || key === 'api_username' || key === 'api_url' || key === 'cache_content')) {
                     chrome.storage.sync.get(key, function(object) {
@@ -791,7 +790,6 @@ YUI.add('bookie-model', function (Y) {
                     });
                 }
             }
->>>>>>> edfa42db8ae4549c502812e53c50c7b70218a365
             //found = localStorage.getItem(key);
         },
 
--- a/bookie/static/js/bookie/model.js
+++ b/bookie/static/js/bookie/model.js
@@ -662,17 +662,13 @@ YUI.add('bookie-model', function (Y) {
          * @private
          *
          */
-        _read: function (options, callback) {
-            this.set('api_url',
-                this._get_data('api_url', this.get('api_url')));
-           this.set('api_username',
-                this._get_data('api_username', this.get('api_username')));
-           this.set('api_key',
-                this._get_data('api_key', this.get('api_key')));
-           this.set('cache_content',
-                this._get_data('cache_content', this.get('cache_content')));
-           this.set('last_bmark',
-                this._get_data('last_bmark', this.get('last_bmark')));
+        _read: function(options, callback) {
+            this._get_data('api_url', this.get('api_url'));
+            this._get_data('api_username', this.get('api_username'));
+            this._get_data('api_key', this.get('api_key'));
+            this._get_data('cache_content', this.get('cache_content'));
+            this._get_data('sync_config', this.get('sync_config'));
+            this._get_data('last_bmark', this.get('last_bmark'));
         },
 
 
@@ -686,12 +682,42 @@ YUI.add('bookie-model', function (Y) {
          * @private
          *
          */
-        _update: function (options, callback) {
-            localStorage.setItem('api_url', this.get('api_url'));
-            localStorage.setItem('api_username', this.get('api_username'));
-            localStorage.setItem('api_key', this.get('api_key'));
-            localStorage.setItem('cache_content', this.get('cache_content'));
-            localStorage.setItem('last_bmark', this.get('last_bmark'));
+        _update: function(options, callback) {
+
+            // everything except last_bmark is synced based on the sync_config flag
+            // no error checking is done at this moment.
+            // check for runtime.lastError if writing to chrome.storage fails
+            // get the sync_config checkbox value
+            // before updating the local storage, the value is updated in the memory
+
+            var sync = this.get('sync_config');
+
+            if (sync) {
+                chrome.storage.sync.set({
+                    'api_url': this.get('api_url'),
+                    'api_username': this.get('api_username'),
+                    'api_key': this.get('api_key'),
+                    'cache_content': this.get('cache_content')
+                });
+            } else {
+                chrome.storage.local.set({
+                    'api_url': this.get('api_url'),
+                    'api_username': this.get('api_username'),
+                    'api_key': this.get('api_key'),
+                    'cache_content': this.get('cache_content')
+                });
+            }
+
+            chrome.storage.local.set({
+                'last_bmark': this.get('last_bmark'),
+                'sync_config': sync
+            });
+
+            // localStorage.setItem('api_url', this.get('api_url'));
+            // localStorage.setItem('api_username', this.get('api_username'));
+            // localStorage.setItem('api_key', this.get('api_key'));
+            // localStorage.setItem('cache_content', this.get('cache_content'));
+            // localStorage.setItem('last_bmark', this.get('last_bmark'));
         },
 
         /**
@@ -704,15 +730,42 @@ YUI.add('bookie-model', function (Y) {
          * @private
          *
          */
-        _get_data: function (key, def) {
+        _get_data: function(key, def) {
+
             var ret,
-                found = localStorage.getItem(key);
-            if (found === null) {
-                ret = def;
-            } else {
-                ret = found;
+                found,
+                sync = this.get('sync_config'),
+                that = this;
+
+            // save the THIS object so that you can update it
+            // later in the async call returns
+
+            function update(object) {
+                if (object.hasOwnProperty(key)) {
+                    found = object[key];
+                } else {
+                    found = null;
+                }
+
+                if (found === null || found === undefined) {
+                    ret = def;
+                } else {
+                    ret = found;
+                }
+                that.set(key, ret);
             }
-            return ret;
+
+            if (sync && (key === 'api_key' || key === 'api_username' || key === 'api_url' || key === 'cache_content')) {
+                chrome.storage.sync.get(key, function(object) {
+                    update(object);
+                });
+            } else {
+                chrome.storage.local.get(key, function(object) {
+                    update(object);
+                });
+            }
+
+            //found = localStorage.getItem(key);
         },
 
         get_apicfg: function () {
@@ -825,7 +878,22 @@ YUI.add('bookie-model', function (Y) {
              *
              */
             cache_content: {
-                value: 'true'
+                value: true
+            },
+
+            /**
+             * suryatech: Since we are using a chrome extension, it makes sense
+             * to use chrome.storage instead of HTML5 localStorage.
+             * You can store objects in it and that should solve the issue of
+             * storing boolean variables as strings.
+             *
+             * @attribute sync_config
+             * @default true
+             * @type boolean
+             *
+             */
+            sync_config: {
+                value: true
             },
 
             /**
--- a/bookie/static/js/bookie/model.js
+++ b/bookie/static/js/bookie/model.js
@@ -741,9 +741,6 @@ YUI.add('bookie-model', function (Y) {
                 that = this,
                 sync;
 
-            // save the THIS object so that you can update it
-            // later in the async call returns
-
             chrome.storage.local.get('sync_config',function(obj){
                 sync = obj['sync_config'];
                 onSyncRead();
@@ -763,8 +760,10 @@ YUI.add('bookie-model', function (Y) {
                 }
                 that.set(key, ret);
 
-                if(callback)
+                if(callback){
                     callback();
+                }
+                    
             }
 
             function onSyncRead(){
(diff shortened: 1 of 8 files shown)


bookie/static/js/bookie/model.js
```diff
@@ -662,17 +662,13 @@ YUI.add('bookie-model', function (Y) {
          * @private
          *
          */
-        _read: function (options, callback) {
-            this.set('api_url',
```
mitechie, line 666: why did you remove the set calls to update the YUI object ATTRS?

bookie/static/js/bookie/model.js
```diff
@@ -686,12 +682,42 @@ YUI.add('bookie-model', function (Y) {
...
+                    'cache_content': this.get('cache_content')
+                });
+            }
+
+            chrome.storage.local.set({
+                'last_bmark': this.get('last_bmark'),
+                'sync_config': sync
+            });
+
+            // localStorage.setItem('api_url', this.get('api_url'));
```
mitechie: so should we set localStorage if sync is not enabled? Or can these commented out lines be removed?

bookie/static/js/bookie/model.js
suryatech, line 666: _get_data function takes care of settings those values.

bookie/static/js/bookie/model.js
```diff
@@ -704,15 +730,42 @@ YUI.add('bookie-model', function (Y) {
...
-            if (found === null) {
-                ret = def;
+                found,
+                sync = this.get('sync_config'),
+                that = this;
+
+            // save the THIS object so that you can update it
+            // later in the async call returns
+
+            function update(object) {
```
mitechie: can you comment on this update method as a bit of a docstring?

bookie/static/js/bookie/model.js
suryatech: If sync is not enabled, we revert back to basic chrome.storage.local
instead of sync. Yes they can be.

bookie/static/js/bookie/model.js
```diff
@@ -662,17 +662,13 @@ YUI.add('bookie-model', function (Y) {
...
-                this._get_data('cache_content', this.get('cache_content')));
-           this.set('last_bmark',
-                this._get_data('last_bmark', this.get('last_bmark')));
+        _read: function(options, callback) {
+            this._get_data('api_url', this.get('api_url'));
+            this._get_data('api_username', this.get('api_username'));
+            this._get_data('api_key', this.get('api_key'));
+            this._get_data('cache_content', this.get('cache_content'));
+            this._get_data('sync_config', this.get('sync_config'));
+            this._get_data('last_bmark', this.get('last_bmark'), callback);
```
mitechie: what's the callback here on the last one?

bookie/static/js/bookie/model.js
```diff
@@ -686,33 +682,115 @@ YUI.add('bookie-model', function (Y) {
...
+            chrome.storage.local.set({
+                'last_bmark': this.get('last_bmark'),
+                'sync_config': sync
+            });
+
+            // localStorage.setItem('api_url', this.get('api_url'));
+            // localStorage.setItem('api_username', this.get('api_username'));
+            // localStorage.setItem('api_key', this.get('api_key'));
+            // localStorage.setItem('cache_content', this.get('cache_content'));
+            // localStorage.setItem('last_bmark', this.get('last_bmark'));
```
mitechie: please remove the commented out lines then please

bookie/static/js/bookie/model.js
```diff
@@ -686,33 +682,115 @@ YUI.add('bookie-model', function (Y) {
...
+                }
+
+                if (found === null || found === undefined) {
+                    ret = def;
+                } else {
+                    ret = found;
+                }
+                that.set(key, ret);
+
+                if(callback)
```
mitechie: always use {} around if blocks please

bookie/static/js/bookie/model.js
```diff
@@ -686,33 +682,115 @@ YUI.add('bookie-model', function (Y) {
...
          *
          * @method _get_data
          * @param {String} key
-         * @param def
+         * @param {String}def
          * @private
          *
          */
-        _get_data: function (key, def) {
+        _get_data: function(key, def, callback) {
```
mitechie: it seems to me that since the loading from chrome storage isn't synchronous that we'd have to do something a bit more explicit to know when the data is read and loaded in. We'd have to start out with default values for each (say null) and as they're read this would have to save the ATTR, and then loop and see if the rest are still null. If so, it carries on, and only when all of them are not null does it callback(). Just sticking callback on the last one does not guarantee it's the last one loaded.

bookie/static/js/bookie/model.js
```diff
@@ -686,33 +682,115 @@ YUI.add('bookie-model', function (Y) {
...
+                    ret = found;
+                }
+                that.set(key, ret);
+
+                if(callback)
+                    callback();
             }
-            return ret;
+
+            function onSyncRead(){
```
mitechie: please use spaces

```
function onSyncRead () {
```

mitechie: can we add comments to this and note why it's here twice?

bookie/static/js/bookie/model.js
```diff
@@ -825,7 +903,22 @@ YUI.add('bookie-model', function (Y) {
...
             cache_content: {
-                value: 'true'
+                value: true
+            },
+
+            /**
+             * suryatech: Since we are using a chrome extension, it makes sense
+             * to use chrome.storage instead of HTML5 localStorage.
+             * You can store objects in it and that should solve the issue of
+             * storing boolean variables as strings.
```
mitechie, line 906: yay!

bookie/static/js/bookie/model.js
suryatech: Function defined in options.js that renders the options view after all async function calls to chrome.storage have returned. This will ensure that the options page has updated info in it.

extensions/chrome_ext/background.html
```diff
@@ -1,16 +1,16 @@
...
-    <script src="lib/tagcontrol.js"></script>
-    <script src="lib/view.js"></script>
+    <meta charset="UTF-8">
+    <title></title>
+    <script type="text/javascript" src="combo.js"></script>
+    <script type="text/javascript" src="config.js"></script>
+    <script type="text/javascript" src="lib/api.js"></script>
+    <script type="text/javascript" src="lib/hash.js"></script>
+    <script type="text/javascript" src="lib/indicator.js"></script>
+    <script type="text/javascript" src="lib/model.js"></script>
```
mitechie, line 11: cool, thanks for the cleanup

extensions/chrome_ext/chrome.js
```diff
@@ -59,7 +59,9 @@ YUI().add('bookie-chrome', function (Y) {
             var hashid = model.get('hash_id');
 
             model.remove(function () {
-                localStorage.removeItem(hashid);
+                chrome.storage.local.remove(hashid);
+                
+                //localStorage.removeItem(hashid);
```
mitechie: please remove commented out lines

extensions/chrome_ext/chrome.js
```diff
@@ -107,9 +109,19 @@ YUI().add('bookie-chrome', function (Y) {
             // display to the user we're working on it.
             model.save(function (data, request) {
                 // make sure that we store that this is a saved bookmark in
-                // the localStorage index
+                // the localStorage index 
+                // instead of localStorage, use chromeStorage to avoid unnecessary
+                // boolean to string and string to boolean conversions. 
+                // you can save only so much info
+                // chrome.storage / localStorage max size limit is 5MB
```
mitechie: cool, thanks. I will say I did a test of the massive.xml file which is some 16k bookmarks and it was under 2MB of hashes for the whole set. Because the hash_ids are a fixed size and we just store a true/false for the value it's fairly small.

extensions/chrome_ext/chrome.js
```diff
@@ -107,9 +109,19 @@ YUI().add('bookie-chrome', function (Y) {
...
+
                 if (data.bmark.hash_id) {
-                    localStorage.setItem(data.bmark.hash_id, 'true');
+                    var hash = data.bmark.hash_id;
+                    chrome.storage.local.set({
+                        hash: true
+                    });
+
+                    //localStorage.setItem(data.bmark.hash_id, 'true');
```
mitechie: remove commented out line

extensions/chrome_ext/chrome.js
```diff
@@ -371,7 +383,12 @@ YUI().add('bookie-chrome', function (Y) {
                             chrome.tabs.getSelected(null, function (tab) {
                                 // we need to hash this into storage
                                 var hash_id = Y.bookie.Hash.hash_url(tab.url);
-                                localStorage.setItem(hash_id, 'true');
+
+                                chrome.storage.local.set({
+                                    hash_id: true
+                                });
+
+                                //localStorage.setItem(hash_id, 'true');
```
mitechie: +1

extensions/chrome_ext/chrome.js
```diff
@@ -479,11 +496,18 @@ YUI().add('bookie-chrome', function (Y) {
...
+            //localStorage.getItem(Y.bookie.Hash.hash_url(url));
+            
+            chrome.storage.local.get(urlHash, function(object){
+                is_bookmarked = object.urlHash;
+            });
 
             // check if we have this bookmarked
             // if so update the badge text with +
+
             if (is_bookmarked === 'true') {
```
mitechie: shouldn't this if condition live in the callback to the get call?

mitechie:
Thanks for this update it's really nice and looks good. The async-ness of the chrome storage access leads me to asking a couple of questions regarding the layout of code in a couple of places.

If you get a chance, check out jslint or jshint and try that against the JS. I try to use that in my editor to help keep my JS code clean and consistent. It'll help with the issues regarding spaces and such.

bookie/static/js/bookie/model.js
suryatech: Sorry. I will clean this up.

extensions/chrome_ext/chrome.js
suryatech: Yes. Most applications require < 5 MB unless it is doing something which is terribly wrong.

extensions/chrome_ext/chrome.js
suryatech: Wow. Thanks for pointing out.

suryatech:
Sublime does have plugins for linting and beautifying. I will definitely clean this up and strive to come up with better code next time.

Thanks for your time.

mitechie:
When I load up the extension for the first time I don't have the default values in the form at all. I filled them all in, but when I hit 'save' it reloaded the window and I lost them all. I'll guess that the form submit action needs to be cancelled so that it submits the form ping and such via ajax vs a normal form submission. 

I'm not able to test out the rest of it right now due to these changes. Can you let me know if you can duplicate this? If not, I can try to chase down what I've got that is wrong/different.

suryatech:
I can't reproduce any of the issues you mentioned. 
1. Default values are filled in. 
2. Window doesn't reload and the AJAX call is made and message is displayed accordingly. 

It appears to work fine at my end.

mitechie:
Found the error in the background page as you suggested.

```
Error in response to storage.get: ReferenceError: webkitNotifications is not defined
    at h.ns.BackgroundPage.Y.Base.create.init_background (chrome-   extension://jphbhcmebeonggghcabcgdeeiidpgmoa/chrome.js:518:34)
    at chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/background.js:3:7
    at q._notify (chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/combo.js:3:129)
    at chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/combo.js:2:5976
    at q._notify (chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/combo.js:3:129)
    at H (chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/combo.js:3:1099)
   at Object.d.Loader._finish (chrome-extension://jphbhcmebeonggghcabcgdeeiidpgmoa/combo.js:91:715) 
```

mitechie:
I no longer have the background page error but I still have issues. When I load the extension it opens the options but all of the fields are empty, no default values. 

I then go through and set them and when I hit save, they clear. There's no errors in the console of the extension or the background page. The only info in the console is a couple of outputs of "on updated"

extensions/chrome_ext/options.js
```diff
@@ -1,8 +1,10 @@
-YUI().use('bookie-model', 'bookie-view', function (Y) {
+YUI().use('bookie-model', 'bookie-view', function(Y) {
     var options = new Y.bookie.OptionsModel();
-    options.load();
-    var view = new Y.bookie.OptionsView({
-        model: options
-    });
-    view.render();
-});
+    options.load(onOptionsLoad);
```
mitechie, line 3: I had to move this down below the function definition to get things to work. The onload event isn't firing and the view isn't created so the form event (that submits and reloads the page) isn't cancelled.

mitechie, line 3: This isn't working out. I'm tempted to punt on this until the chrome extension has it's own repo and perhaps we ditch some of the YUI view/sync stuff.

mitechie:
After talking this over I'm going to close this pull request. We'll be able to reference the work later, but we have some pre-work to port the chrome extension to raw JS and move it to its own repository.

PR #346, opened 2014-03-10
check if api keys match

rakeshmahadasa:


rakeshmahadasa:
Sorry for the unnecessary commits , overlooked them.
https://github.com/infinitum-rakesh/Bookie/commit/1822777614a900cbfb98f6cc73d4565e2154ee1d
This is my latest commit to check the authenticity of api keys.

bookie/views/api.py
```diff
@@ -63,10 +63,20 @@ def ping(request):
     """Verify that you've setup your api correctly and verified
 
     """
-    return _api_response(request, {
-        'success': True,
-        'message': 'Looks good'
-    })
+    params = request.params
+    username = params.get('username', None)
```
mitechie, line 67: the username should be in the request.use.username

bookie/views/api.py
```diff
@@ -63,10 +63,20 @@ def ping(request):
...
+    api_key = params.get('api_key', None)
+    user = UserMgr.get(username = username)
```
mitechie, line 69: no space between the uesrname=username in this case. Make sure to run 'make lint' and check for any errors in the stuff you've added.

bookie/views/api.py
```diff
@@ -63,10 +63,20 @@ def ping(request):
...
+    if api_key == user.api_key:
+        return _api_response(request, {
+            'success': True,
+            'message': 'Looks good'
+        })
+    else:
+        return _api_response(request, {
+            'success': False,
+            'message': 'API keys does not match.'
```
mitechie, line 78: So the user just submitted one api key. Let's tell them that their api_key is not correct. I think that'll be clearer to users vs a match issue.

bookie/views/auth.py
```diff
@@ -194,7 +194,18 @@ def reset(request):
         new_username = params.get('new_username', None)
         error = None
 
-        if not UserMgr.acceptable_password(password):
+        # Check whether username exists or not.  During signup request , a
```
mitechie, line 197: I thought this already landed in trunk? 

https://github.com/bookieio/Bookie/blob/develop/bookie/views/auth.py#L197

Can you pull this out of your pull request?

mitechie:
This looks really good! Thanks so much for the branch and the great test. I've added a few comments for some small nitpicks but look forward to landing this. Make sure to add your name to the credits file in your pull request if you're not already listed.

bookie/views/api.py
rakeshmahadasa, line 67: Oh seems like i did not pull the final update. yes params only contains api_key and nothing else so we have to get the username from request. I did that and will pull that now. thanq.

bookie/views/auth.py
rakeshmahadasa, line 197: sure.. overlooked that sry..

mitechie:
Closing, new pull request was created.

PR #348, opened 2014-03-11
Api key validation on ping request with test added

rakeshmahadasa:


mitechie:
Thanks so much for this bug fix. It's landed and should be deployed later tonight.

PR #351, opened 2014-03-12
Made Makefile Arch and Fedora compatible

### LINKED ISSUE #310: Make Bookie archlinux compatible (or even distro independent if possible)

sanchitgangwar:
Fixes #310.


dyladan:
/scripts/js/jsmin_all.py depends on the system python which is python3 by default in arch. Should be `#1/usr/bin/env python2`.

In Arch services do not start after installing. You need to start redis.

I am working on these changes now

PR #353, opened 2014-03-12
fetching popular bookmarks based on clicks

kshitijiiit:
- Fixed API for fetching popular bookmarks.
- Added views for popular bookmarks.
- Option for viewing popular bookmarks (both personal and all) in the footer
- Added API tests

### DIFF
--- a/bookie/static/js/bookie/api.js
+++ b/bookie/static/js/bookie/api.js
@@ -329,6 +329,7 @@ YUI.add('bookie-api', function (Y) {
             data: {
                 count: 10,
                 page: 0,
+                sort: 'popular',
                 with_content: false
             }
         }, {
--- a/bookie/static/js/bookie/api.js
+++ b/bookie/static/js/bookie/api.js
@@ -329,7 +329,6 @@ YUI.add('bookie-api', function (Y) {
             data: {
                 count: 10,
                 page: 0,
-                sort: 'popular',
                 with_content: false
             }
         }, {
--- a/bookie/views/api.py
+++ b/bookie/views/api.py
@@ -63,10 +63,23 @@ def ping(request):
     """Verify that you've setup your api correctly and verified
 
     """
-    return _api_response(request, {
-        'success': True,
-        'message': 'Looks good'
-    })
+    rdict = request.matchdict
+    params = request.params
+    username = rdict.get('username', None)
+    api_key = params.get('api_key', None)
+    user = UserMgr.get(username=username)
+    
+    # Check if user provided the correct api_key
+    if api_key == user.api_key:
+        return _api_response(request, {
+            'success': True,
+            'message': 'Looks good'
+        })
+    else:
+        return _api_response(request, {
+            'success': False,
+            'message': 'API key is invalid.'
+        })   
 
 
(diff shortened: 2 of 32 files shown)


bookie/views/api.py
```diff
@@ -393,8 +400,6 @@ def user_bmark_count(request):
     })
 
 
-@view_config(route_name="api_bmarks_popular", renderer="jsonp")
-@view_config(route_name="api_bmarks_popular_user", renderer="jsonp")
 @api_auth('api_key', UserMgr.get, anon=True)
```
mitechie: we should just remove this method then since it's no longer part of the api.

bookie/views/api.py
```diff
@@ -306,6 +306,13 @@ def bmark_recent(request, with_content=False):
     page = int(params.get('page', '0'))
     count = int(params.get('count', RESULTS_MAX))
 
+    # check if we have a ordering by popularity specified
```
mitechie: please update this to be a complete sentence with a capital letter and ending in a period.

bookie/static/js/bookie/api.js
```diff
@@ -329,6 +329,7 @@ YUI.add('bookie-api', function (Y) {
             data: {
                 count: 10,
                 page: 0,
+                sort: 'popular',
```
mitechie: we don't want to default to sorting by popular

bookie/views/api.py
```diff
@@ -306,6 +306,13 @@ def bmark_recent(request, with_content=False):
...
+    order_by = params.get('sort', None)
+    if order_by == "popular":
+        order_by = Bmark.clicks.desc()
```
mitechie, line 325: Do we have click values? Should this instead be Bmark.hashed.click.desc()?

Remember we've got two click counts. one per user and one for the url as a whole?

mitechie:
So the change looks ok on the surface. I'd check the database and make sure you have real click values to sort on.

The other thing you can do to help debug is to make manual api requests and see if you get different results against your test system.

I use a tool called httpie, which is written in python and provides a pretty output. https://pypi.python.org/pypi/httpie/0.8.0

You can then use that to manually build a request to the api and add the flag and see how your results change or not.

bookie/views/api.py
kshitijiiit: Yes, I'll do that in the next commit.

bookie/views/api.py
kshitijiiit: Sure.

bookie/static/js/bookie/api.js
kshitijiiit: Yes, I know but I was just trying this out to see if it works. And it doesn't seem to reflect the changes in the final view. Would you know the reason for it? 
The api seems to be working absolutely fine but I really can't find out how to reflect that change in the view of "/recent".

bookie/views/api.py
kshitijiiit, line 325: Bmark.hashed.click is the global click count for a given url. I think we would need to sort by local clicks for each user in this api which is given by Bmark.clicks. Perhaps a different call can be made to sort by the global count. What do you think?

mitechie:
Thanks so much for this change. I've landed it and will update it on the live site soon.

Issue #71, opened 2011-05-19
import from Firefox dump


mitechie:
Tweaked the google import to work for a Firefox html dump. It doesn't dump tags though. The <h3> is the directory of the bookmark. The backup/json dump has the info, but it's very tied to the db table relation.

Basically, anything of type 2 is tags. It's nested and you have to sort.

mitechie:
http://stackoverflow.com/questions/464516/firefox-bookmarks-sqlite-structure

describes the structure a bit

silopolis:
This is maybe the feature I miss most but may be I can upload to delicious or google and then import in bookie... Btw, delicious behaviour of converting folders to tags is cool for people like me that took years before understanding the power of tagging ;-)

mitechie:
If you have a firefox bookmarks json dump you don't mind sharing I'd love to get a copy and we can look at adding an importer. It's usually not terribly hard. We just need to write a custom importer class here:

https://github.com/mitechie/Bookie/blob/develop/bookie/lib/importer.py

mitechie:
Amazingly updated and on the live site. Woot!

PR #355, opened 2014-03-13
issue #71

preddy5:


bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
         # fetch its content.
         for bid in ids:
             tasks.fetch_bmark_content.delay(bid)
+
```
mitechie, line 447: two space between classes, make sure to run make lint and correct any new issues.

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+    @staticmethod
+    def _is_firefox_format(json, fbookmark_type, can_handle):
+        """Verify that this import file is in the firefox backup
+        export format
+
+        Firefox json file has a variable "type" which is equal to
+        "text/x-moz-place-container"
+        """
+        if (json['type'] == fbookmark_type):
```
mitechie: no need for the (). Just if json['type']...

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+        In order to check the file we have to read it and check it's content
+        has a variable "type" which is equal to "text/x-moz-place-container"
+        """
+        if (file_io.closed):
+            file_io = open(file_io.name)
+        file_io.seek(0)
+        can_handle = False
+        try:
+            backup_json = json.load(file_io)
+        except:
```
mitechie, line 479: should this reset the file stream like we do at the end of the can_handle method?

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+                                                         can_handle)
+
+        # make sure we reset the file_io object so that we can use it again
+        file_io.seek(0)
+        return can_handle
+
+    def process(self):
+        """Process an json firefox bookmarks export and import them into bookie
+        """
+        MOZ_CONTAINER = "text/x-moz-place-container"
```
mitechie: this is duplicated. I'd move this to be a module level constant then outside of this function.

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+        # HACK: Firefox' JSON writer leaves a trailing comma
+        # HACK: at the end of the array, which no parser accepts
+        if content.endswith(u"}]},]}"):
+            content = content[:-6] + u"}]}]}"
+        root = json.loads(content)
+
+        # make a dictionary of unique bookmarks
+        bmap = {}
+
+        def bmap_add(bmark, bmap):
```
mitechie: why not just stick this class on the class? self.bmark_app(bmark, bmap)?

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+            if bmark["uri"] not in bmap:
+                bmap[bmark["uri"]] = bmark
```
mitechie: if it is in the map can they have different tags? Maybe the importer should combine all the tags on any found versions of that uri?

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+
+        def is_good(ch):
```
mitechie: ch? can we spell this out to be more clear?

mitechie: and add a comment on what the split/etc is checking?

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+            return not ch["uri"].split(":", 1)[0] in UNWANTED_SCHEME
+
+        # find toplevel subfolders and tag folders
+        catalogs = []
```
mitechie: you comment this as subfolders and tag folders but name the variables catalogs. Let's pick the one that's most clear and keep it consistent.

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+            for bmark in tag["children"]:
+                if bmark["type"] == MOZ_PLACE and bmark.get("uri") and is_good(bmark):
+                    bmap_add(bmark, bmap)
+                    if not "tags" in bmap[bmark["uri"]]:
+                        bmap[bmark["uri"]]["tags"] = []
+                    bmap[bmark["uri"]]["tags"].append(tag["title"].replace(" ", "-"))
+
+        # save the bookmarks
+        ids = []
+        for url,metadata in bmap.items():
```
mitechie: space between the , and m

bookie/lib/importer.py
```diff
@@ -440,3 +444,142 @@ def process(self):
...
+            if metadata.get('annos') is not None:
```
mitechie, line 554: comment on what 'annos' is in metadata and why we care about it?

bookie/tests/test_utils/test_imports.py
```diff
@@ -407,6 +437,49 @@ def test_import_process(self):
         # now let's do some db sanity checks
         self._chrome_data_test()
```
mitechie, line 439: two spaces here please

mitechie:
This is a great patch. A lot of people will be excited to be able to load their Firefox bookmarks into Bookie.

The code looks really good. I've noted a few small things that I'd appreciate it if you'd consider making it a bit more clear or tweaking.

The test is a really really good test. There's some special logic in the importer for dealing with nested folders. Can we make sure the test exercises that code and has several folders deep to make sure that it's not broken in the future?

mitechie:
Thanks so much for this commit! This is really great and I've landed it. I'll update the site shortly.

PR #352, opened 2014-03-12
Notify user that a duplicate tag has been entered

suryatech:
Fixes #198. 

Fade out the original tag and retain the recently entered tag

PR #357, opened 2014-03-14
Duplicate tags

suryatech:
New pull request. Old request: #352.


bookie/static/js/tests/test_tagcontrol.js
```diff
@@ -177,6 +177,86 @@ YUI.add('bookie-test-tagcontrol', function (Y) {
...
+
+            // Value should be test test1 test2 test2 immediately after we hit space.
+            Y.Assert.areEqual(
+                'test test1 test2 test2',
+                this.t.get('value'),
+                "The input should have four tags as values: " + this.t.get('value')
+            );
+
+            function test1(){
+                var value = this.t.get('value');
```
mitechie: This trick with a callback like this is to use the wait and resume test helpers to make sure the test fails if this callback is not called. 

http://yuilibrary.com/yui/docs/test/#asynctests

If this races to the end it might mark the test as passing without hitting the callback at all.

Issue #349, opened 2014-03-11
Search against keyword with # symbol


puntonim:
Searching against the keyword "#fedora" returns a page with the spinner image rotating forever
This happens at bmark.us:
https://bmark.us/results?search=%23fedora&content=true&submit=Search
but it seems not to happen in our local devs

puntonim:
We can find clues in the Javascript console
I'm working on that

mitechie:
Thanks landed.

PR #360, opened 2014-03-14
Fix issue #349 - Search keywords with hash symbol

puntonim:
The issue was related to the API. A search with keyword `#fedora` must be escaped to:

```
/api/v1/bmarks/search/%23fedora
```

The search page in Bookie website makes an API call without escaping keywords.
We could have solved by changing the API (which has a big impact) or by changing the JS code in the search webpage to introduce the escaping of keywords. I chose the second option.


mitechie:
Thanks so much for the nice fix. I've landed it and will update the live site soon.

Issue #198, opened 2012-12-13
chrome extension magically removes duplicate tags


widox:
When adding tags to a new bookmark via the Chrome extension, and the same tag already exists, it just removes the duplicate tag I just typed with no warning why.

Example typing:
tag1 <space> tag2 <space> tag1 <space> at this point the second tag1 just gets deleted.

It would be nice for some visual indication as to why it just got removed. e.g. blink the first tag with the same name

PR #361, opened 2014-03-15
Added test for duplicate tags. Added animation.

suryatech:
Modified tests to prevent race situations. Used wait construct instead.

### DIFF
--- a/bookie/static/js/bookie/tagcontrol.js
+++ b/bookie/static/js/bookie/tagcontrol.js
@@ -185,6 +185,8 @@ YUI.add('bookie-tagcontrol', function (Y) {
          * making sure we're updating our srcNode with the list of tags for
          * form submitting later on.
          *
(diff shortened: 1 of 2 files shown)


sambuddhabasu:
Tested the pull request. The JS tests run fine. make lint is clean.
@mitechie Do let me know if I am missing any other step for testing the code.

suryatech:
@sammyshj : Thanks for your time. What do you think about the animation. Is 0.1 s fine or should it be increased/decreased?

mitechie:
Thanks for this, since the original issue in #198 was that Bookie removed the dupe and it wasn't clear why, I slowed it down to .4s so that it's obvious that there's something going on and the user hopefully puts it all together now. 

Thanks for the branch, the tests, and your patience as it all got worked out.

suryatech:
Great! I thought that a larger duration might make it 'appear' slow. Anyways, thanks for the change.

PR #362, opened 2014-03-15
Weekly bookmark count with tests

sambuddhabasu:
For bookmark count, when time period > 31, a weekly bookmark count response is now returned. Tests also included.


sambuddhabasu:
Pull request updated at #364

Issue #88, opened 2011-06-20
update api to provide stats information


mitechie:
Might be cool to provide a /stats api url for things like number of bookmarks, tags, etc in the last seven days and what not. Then the mobile ui could show some nice trends/activity info. I know I'd like it, but other users might as well and it's all non-personal data so should be good to share via api.

sambuddhabasu:
I would like to work on this issue.
Already the /dashborad api url shows number of bookmarks and number of unique urls. So, the other things could be added to the same api.

PR #364, opened 2014-03-16
Weekly bookmark count with tests

sambuddhabasu:
For bookmark count, when time period > 31, a weekly bookmark count response is now returned. Tests also included.

### DIFF
--- a/bookie/models/stats.py
+++ b/bookie/models/stats.py
@@ -139,12 +139,20 @@ class StatBookmarkMgr(object):
                 end_date = start_date + timedelta(days=STATS_WINDOW)
         # Since we're comparing dates with 00:00:00 we need to add one day and
         # do <=.
-        return [
+        response = [
             StatBookmarkMgr.get_user_bmark_count(
                 username, start_date, end_date + timedelta(days=1)),
             start_date,
             end_date
         ]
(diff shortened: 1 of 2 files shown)


mitechie:
Thanks so much for this. It's got some good ideas, but there's a few disadvantages I can forsee. 

The model code should always be minimal and keep to proving the real data without any (or as little) application logic as possible. This change makes it impossible to get the stats numbers for a whole year. This limits our ability to do future things. I think a better thing to do would be to add an argument to request data by week, by day, by month. And adjust the queries done to support those. 

Heading down this path means that the View would need to adjust the stats loading method call to supply the range. It could look at the date range supplied, and auto adjust to weeks/months based on the UI input. The only issue here is that, we again limit what data can be fetched. So the API should respect the user's wishes. 

So in order to support this auto adjusting idea, without limiting the future use of the api and data, would be to move the logic into the JS code that makes the API request. This means that only the Bookie user interface is limited, and if others can find great new ways to use the API they'll have the ability to do so. 

What do you think?

sambuddhabasu:
The idea of adding an argument to request data by week, by day, by month seems fine. Would look closely into this right after I'm done with issue #88

Issue #363, opened 2014-03-16
The Chrome extension couldn't work


jerryleooo:
When I click the 'Save' button of the Bookie Chrome extension, there just is a circle progressbar rolling all the time and when I logined on the bmark.us I found nothing.
I have set both the username and the api key.

suryatech:
A quick peek suggests that something has gone wrong in the latest release. Will let you know what it is.

mitechie:
Thanks for the report. I see the error and will have a fix up shortly. It'll take a little bit for the chrome store to process the update.

mitechie:
I've uploaded version 0.4.16 which corrects this. Once the Chrome store processes the update it should auto upgrade everyone's extension. 

Note: you can try to force the upgrade from the "manage" section of the extensions. It has been taking 30-40minutes for the chrome web store to process new versions lately. 

Closing, but if you don't see a fix in the next hour or so please reopen this issue.

jerryleooo:
Version 0.4.16 has fixed this issue, I tried it just now.

PR #365, opened 2014-03-16
Fixes issue #363.

suryatech:
Misses the safety check. When you are about to bookmark something for the first time, the request sent to the server contains the stored field by invoking model.get('stored'). Unfortunately, it does not account for the undefined state of the variable and performs operations on it, which results in an unexpected error.

The server replies back with a Bmark object containing the "stored" variable which is used in future communications.


mitechie:
Thanks for the update. Sorry I missed you were looking into it. I've pushed an update to the chrome store. Closing this branch.

PR #367, opened 2014-03-16
Delete non activated accounts. Fixed #234

### LINKED ISSUE #234: non-activated accounts should be cleaned up occassionally
When people signup it's a two step process. First, they submit an email address and we email them an activation email where they can set a username/password. Many spam accounts and even real people don't complete the second step. We should clean these out ever so often. Let's say, if after 30 days you've not activated, user and track the failed activation for stats purposes.

sambuddhabasu:


bookie/models/auth.py
```diff
@@ -133,6 +136,20 @@ def count():
...
 
     @staticmethod
+    def delete_non_activated_account():
+        """Delete user accounts which are not verified since
+        30 days of signup"""
+        test_date = datetime.utcnow() - NON_ACTIVATION_AGE
+        qry = User.query.filter(User.activated == False).\
+            filter(User.invited_by == u'signup').\
+            filter(User.last_login == None).\
+            filter(User.signup < test_date).all()
```
mitechie: I think this should be against the activation date, not the signup date.

bookie/models/auth.py
```diff
@@ -133,6 +136,20 @@ def count():
...
+        for account in qry:
+            qry_users = User.query.filter(User.id == account.id).delete()
```
mitechie: this can be one query.

```
User.query.join(User.activation).options(eager_load(User.activation)) 
```

or something like that. Check the other queries for joined examples.

bookie/tests/test_auth/test_model.py
```diff
@@ -107,6 +111,33 @@ def test_activation_cascade(self):
...
+        self.assertEqual(
+            1,
+            len(activations),
+            'We should have a total of 1 activation: ' + str(len(activations)))
+        self.assertEqual(
+            2,
+            len(users),
+            'We should have a total of 2 users: ' + str(len(users)))
+        activations[0].valid_until = datetime.utcnow() - timedelta(days=35)
+        UserMgr.delete_non_activated_account()
```
mitechie: in order to be a good test we should have both positive and negative test data. In this case, we want to make sure that if we have a recent valid signup it's not deleted as part of this work. We should also include a user that is not in signup mode, but is activated, and possible at least one user that's a logged in user that's forgotten their password. This makes us certain that the code works as intended and will not accidentally remove legit users from the system. 

Sorry for the extra test work, but this is a particularly sensitive bit of code that can cause a nightmare to fix if it goes wrong.

sambuddhabasu:
Ran make lint and rebased the previous commits.

mitechie:
Thanks so much for the test update. I hope you don't mind, but would you consider adding a new admin only api endpoint for this that has a 'dry-run' feature?

The general idea is this: 

I want to run this but I'd like to have it first tell me who all it plans on deleting. If the .delete() is turned into a .all() it can return all those users. The api endpoint could be hit to do the deletion, but with a flag on it, could request to see the list vs doing the deleting. 

The dry-run version should  be the initial endpoint so that it's safe to hit the url without changing anything.

Maybe something like 
    api/v1/a/garden-activations
    api/v1/a/garden-activations/go

What do you think?

sambuddhabasu:
Yeah, this looks like a pretty solid plan. This way, we could get an idea of the different users that are currently having a non-activated account. Passing an argument, will delete all of those accounts.
The updated code will be available shortly.

sambuddhabasu:
All the commits have been squashed and rebased. Tests included for api calls and deletion of non activated accounts. make lint is now clean. Fixed #234

bookie/routes.py
```diff
@@ -214,6 +214,16 @@ def build_routes(config):
         "/api/v1/a/applog/list",
         request_method="GET")
 
+    config.add_route(
+        "api_admin_non_activated",
+        "/api/v1/a/nonactivated",
```
mitechie, line 219: perfect

bookie/routes.py
```diff
@@ -214,6 +214,16 @@ def build_routes(config):
...
+        request_method="GET")
+
+    config.add_route(
+        "api_admin_delete_non_activated",
+        "/api/v1/a/nonactivated",
```
mitechie, line 224: exactly, one url but with different methods

bookie/models/auth.py
```diff
@@ -133,6 +137,27 @@ def count():
...
+        test_date = datetime.utcnow() - NON_ACTIVATION_AGE
+        qry = User.query.join(User.activation).\
+            options(contains_eager(User.activation)).\
+            filter(User.id == Activation.id).\
+            filter(User.activated == False).\
+            filter(User.last_login == None).\
+            filter(Activation.valid_until < test_date).all()
+        # Delete the non activated accounts only if it is asked to.
+        if delete:
+            for account in qry:
```
mitechie: you're looping 'for account in qry'? Why not just do the query again but with a .delete vs a .all?

So I'd do something like

```
qry = User.query.join(User.activation).\
           options(contains_eager(User.activation)).\
          filter(User.id == Activation.id).\
          filter(User.activated == False).\
          filter(User.last_login == None).\
          filter(Activation.valid_until < test_date)

if not delete:
    return qry.all()
else:
    qry.delete()
```

mitechie:
Tested this out with an old backup copy of the live site and it worked well other than a small tweak to the api response needed. Thanks so much for the hard work and patience while this get landed. I'll work on getting it on the live site and running it there to clean up accounts soon.

Issue #266, opened 2014-01-24
Signup process needs some love 


mitechie:
It still talks about being in alpha, needing an invite, etc. It's not friendly for new users.

aka001:
I want to work on this issue. Kindly describe what should we exactly do in this ticket?

mitechie:
Some of it is textual. 
- Bookie isn't alpha, that note should be pulled out
- You don't need an invite to get into bookie any more. That should be pulled out
- Once you click the activate link in your email and set a password, bookie should auto log you in and show you the page for your bookmarks.

badanomaly:
If nobody is working on this issue I would like to start working on it.

aka001:
yaa continue on it. :)

rakeshmahadasa:
There are problems with signup. I dont know whether Im the only one facing the problem 
1. Confirmation link is not being delivered 

2.Even some valid email addresses like ( coders.infinitum@gmail.com ) are marked as invalid as an invalid email error is being generated when trying to use forgot password option

mitechie:
On Tue, 04 Mar 2014, Rakesh Mahadasa wrote:

> There are problems with signup. I dont know whether Im the only one facing the problem
> 1. Confirmation link is not being delivered
>    2.Even some valid email addresses like ( coders.infinitum@gmail.com ) are
>    marked as invalid as an invalid email error is being generated when
>    trying to use forgot password option

Is this in a local environment? You have to run the 'make smtp' to runa
fake email server. That output needs to be base64 decoded to view the
email.

https://github.com/bookieio/Bookie/blob/develop/bookie/views/api.py#L713

Shows the code for checking the email. Can you note what message
specifically you're hitting?

## 

Rick Harding
@mitechie
http://blog.mitechie.com
http://lococast.net

rakeshmahadasa:
Sorry My bad Im submitting "," instead of "."
Now with the email part corrected Its working as expected
but still Im not getting the conformation email. I did signup using two emails till now 
1.infintum.rakesh@gmail.com
2.coders.infinitum@gmail.com

I didn't get conformation emails. Im not using local environment. 
I faced the same problem initially ( rakeshnitcalicut@gmail.com ) and then you manually deleted my entry and asked me to signup again and I got conformation mail from your email address.

mitechie:
On Tue, 04 Mar 2014, Rakesh Mahadasa wrote:

> Sorry My bad Im submitting "," instead of "." with the email part corrected Its working as expected
[...]

PR #369, opened 2014-03-19
issue #266

preddy5:


preddy5:
"Some of it is textual.
- Bookie isn't alpha, that note should be pulled out
- You don't need an invite to get into bookie any more. That should be pulled out
- Once you click the activate link in your email and set a password, bookie should auto log you in and show you the page for your bookmarks."

1,2 done, i think 3 is already in place

preddy5:
@mitechie what more edits are needed?

sambuddhabasu:
I have tested the code. The code seems to run fine. Activation emails are sent as expected with the desired message. make lint is clean.
As far as the three points go, 1 and 2 have been fixed with the current pull request and 3 has been already taken care of before.

mitechie:
Thanks for helping to polish things up and thanks for the QA Sambuddha.

PR #370, opened 2014-03-19
Email server docs added

sambuddhabasu:


preddy5:
hey @sammyshj write docs how to install sendmail to virtualenv of bookie cause it is not present in requirements.txt and is required for smtp

sambuddhabasu:
@pradyu1993 The reason why it is not present in requirements.txt or installed by default is that an SMTP client like msmtp is for development. Already on running 'make smtp', runs a python file which uses python module inbox to append the emails in a file called 'email_log'.

preddy5:
@sammyshj sorry not sendmail "inbox", mention in the docs how to install inbox to the virtualenv cause from past few days people were complaining about make smtp not working though sendmail or msmtp was installed to ubuntu environment

sambuddhabasu:
@pradyu1993 If you look closely, the instructions have already been added to this file
https://github.com/bookieio/Bookie/blob/develop/docs/install.rst
For some reason, the online docs have not been updated, which will be done soon.

preddy5:
@sammyshj ya, i think it should also be added to "Setting Up Email Server" docs think about it, just a opinion

sambuddhabasu:
@pradyu1993 Thanks for that. I will talk about this with @mitechie .

mitechie:
Thanks so much for these changes. I've tweaked them a small bit and landed them. I'll make sure readthedocs updates shortly.

PR #372, opened 2014-03-20
Suggested tags feature implemeted for edit page

rakeshmahadasa:


bookie/tests/test_models/test_tagmgr.py
```diff
@@ -62,3 +62,60 @@ def test_case_insensitive(self):
...
+            'extended': u'',
+            'tags': u'',
+            'api_key': key,
+        }
+        
+        res = self.testapp.post('/api/v1/admin/bmark',
+                                params=test_bmark,
+                                status=200)
+        
+        # generate tags for the bookmark
```
mitechie: from here I'd expect the test to add a set of known content to the bookmark as readable data. Then the test would go to the edit url for the bookmark and make sure that the expected set of suggested tags exist in the UI in the suggested tags section.

bookie/models/__init__.py
```diff
@@ -222,8 +223,18 @@ def complete(prefix, current=None, limit=5, username=None):
 
             return DBSession.execute(query)
 
+
+    @staticmethod
+    def sort_word_strength(tags):
+        size = len(tags)
+        for x in range(0,size-1):
```
mitechie: make sure to run make lint, this should throw an error. There should be spaces after the comma here and in other places as well.

bookie/models/__init__.py
```diff
@@ -222,8 +223,18 @@ def complete(prefix, current=None, limit=5, username=None):
...
+            for y in range(x+1,size):
+                if tags[x][1] < tags[y][1]:
+                    tags[x] , tags[y] = tags[y] , tags[x]
+        return tags
```
mitechie: check out https://wiki.python.org/moin/HowTo/Sorting as I think we can do this sorting more efficiently using the built in sorting tools.

bookie/models/__init__.py
```diff
@@ -222,8 +223,18 @@ def complete(prefix, current=None, limit=5, username=None):
...
+        size = len(tags)
+        for x in range(0,size-1):
+            for y in range(x+1,size):
+                if tags[x][1] < tags[y][1]:
+                    tags[x] , tags[y] = tags[y] , tags[x]
+        return tags            
+
     @staticmethod
-    def suggestions(bmark=None, recent=True, url=None, username=None):
+    def suggestions(bmark=None, recent=True, url=None, username=None,edit=False):
```
mitechie: another space please.

bookie/models/__init__.py
```diff
@@ -234,11 +245,33 @@ def suggestions(bmark=None, recent=True, url=None, username=None):
...
+        # Suggested tags feature only supported for edits. 
+        if not edit:
+            #  If url is None return empty suggested tag list
+            if url is None:
+                return tag_suggest
+            else:
+                bmark = BmarkMgr.get_by_url(url)    
+                # If bmark is not parsed return empty tag list
+                if bmark.readable.status_code == '900':
+                    return tag_suggest
```
mitechie: notice that you don't let the existing code that returns recent tags work at all here. You're returning values. Just because the new tag suggest code cannot suggest, does not mean it should return and prevent later code from attempting to provide some help. 

What this should do is append values to the tag_suggest list if it can, otherwise it should leave it alone and pass through to the code afterwards.

mitechie:
Thanks so much for this branch. This is REALLY exciting as it's the start of a feature I always wanted Bookie to get built in. 

There's a few small tweaks we need to look into before landing. Please let me know if any of the comments are unclear.

bookie/models/__init__.py
rakeshmahadasa: sure , this wont happen again.

bookie/models/__init__.py
rakeshmahadasa: thanx for the link , I thought there is no builtin function which sorts a list of lists based on given key. will update it

bookie/models/__init__.py
rakeshmahadasa: yeah , got it.

bookie/bcelery/tasks.py
```diff
@@ -276,7 +276,7 @@ def fetch_unfetched_bmark_content(ignore_result=True):
 
     url_list = Bmark.query.outerjoin(
         Readable, Bmark.readable).\
-        filter(Readable.imported == None).all()
+        filter(Readable.imported is None).all()
```
mitechie: this has to be imported.is_(None)

bookie/models/__init__.py
```diff
@@ -234,16 +235,42 @@ def suggestions(bmark=None, recent=True, url=None, username=None):
...
-
-            if recent:
-                tag_suggest.extend(recent.tag_str.split(u" "))
-
-        tag_list = list(set(tag_suggest))
-        return tag_list
+            recent_tags.extend(recent.tag_str.split(u" "))
+        recent_tags = list(set(recent_tags))
+        # Suggested tags feature only supported for edits. 
+        if not edit:
```
mitechie: I thought this was going to take place if it was an edit?

rakeshmahadasa: Ah that was a very bad use of variable name... will pull the updated code. :)

bookie/models/__init__.py
```diff
@@ -234,16 +235,42 @@ def suggestions(bmark=None, recent=True, url=None, username=None):
...
+                    return recent_tags
+                else:
+                    content = bmark.readable.content
+                    # Remove unicode character while printing
+                    clean_content = ("".join(BeautifulSoup(content).findAll(text=True)).encode('ascii', 'ignore'))                     
+                    get_tags = extract.TermExtractor()
+                    tag_suggest = get_tags(clean_content)
+                    tag_suggest = sorted(tag_suggest, key=lambda tag_suggest: 
+                                         tag_suggest[1], reverse=True) 
+                    for tag in tag_suggest:
```
mitechie: so you can avoid this loop by doing an http://docs.python.org/2/library/stdtypes.html#set.update (make sure this is a set so you can't get dupes)

bookie/models/__init__.py
```diff
@@ -234,16 +235,42 @@ def suggestions(bmark=None, recent=True, url=None, username=None):
...
+                        tag_list.append((tag[0]))
```
mitechie: so then the tag_suggest should just be update'd onto the recent_tags list from above.

bookie/models/__init__.py
```diff
@@ -234,16 +235,42 @@ def suggestions(bmark=None, recent=True, url=None, username=None):
...
+                    # return maximum of 5 tags
+                    # extend the list with recent tags
+                    tag_list.extend(recent_tags)
+                    if len(tag_list) >= 5:
```
mitechie: Can we make this a constant so that we can tweak it easily in the future if we want 6 or 7 we can tweak it.

mitechie:
This is looking very good. A couple of suggestions. I'll test it out after you peek at my comments.

mitechie:
This is awesome and I've pulled it in. 

Please make sure to check out the hacking doc. I had to do a lot of cleanup because you had a bug fix in place that didn't work around lower casing the username on signup. That should have been a different branch. It appears you're working in the develop branch all the time and the hacking doc will help suggest a better git workflow.

PR #374, opened 2014-03-22
Added Generate new Api Key Feature.

rakeshmahadasa:


bookie/tests/test_webviews/test_account.py
```diff
@@ -67,3 +68,24 @@ def test_delete_all_bookmarks_without_confirmation(self):
             "Delete request not confirmed. Please make sure" +
             " to enter &#39;Delete&#39; to confirm." in res.body,
             msg="Request should contain the appropriate message.")
+    def test_new_api_generation(self):
```
mitechie, line 71: there should be a new line before this test starts

bookie/views/auth.py
```diff
@@ -262,6 +263,27 @@ def reset(request):
             'user': user,
         }
 
+@view_config(route_name="generate_new_api", renderer="/accounts/index.mako")
+def api_key_reset(request):
```
mitechie, line 267: this method doesn't have authentication requirements on it.

mitechie:
This is a good start at the idea. I've got a couple of suggestions.

First, I'd suggest that this feature get added to the api. It should be a POST version of the existing api_key endpoint described at

http://docs.bmark.us/en/latest/api.html#username-api-key

Then the Account page JS should be updated to call that api endpoint in Javascript. When you view your api_key, there should be a button to "Generate new api key". When clicked it would generate a call to that api endpoint created. 

That JavaScript should go into the AccountView here:

https://github.com/bookieio/Bookie/blob/develop/bookie/static/js/bookie/view.js#L1143

It handles all of the JS for the user's account page. 

The final bit is that the button handling code in the JS should talk to a new api key call in the JS api.js file. This link goes to the current api call that fetches a user's api key.

https://github.com/bookieio/Bookie/blob/develop/bookie/static/js/bookie/api.js#L682

rakeshmahadasa:
deleted the branch during a clean up unexpectedly . reopening

rakeshmahadasa:
api end point added #395

PR #378, opened 2014-03-24
fulltext updating script for readable bookmarks

preddy5:


mitechie:
This looks amazingly awesome. Per our chat in irc, since we've had a typo there in the view I'd love if we could write a test that makes sure that a certain url call will generate the right arguments to the fulltext search side.

preddy5:
@mitechie

mitechie:
Thanks for this. I've done a few tweaks to the branch to land. The requirements.txt needed to be updated along with the download-cache to function properly. I'll try to get this on production soon and start running the script. It'll be nice to see if this helps improve fulltext search abilities.

Issue #368, opened 2014-03-17
Users unable to login if their username contains an uppercase letter


HubertChen:
Currently, the login page automatically lowercases all letters in usernames. This causes conflict when a user tries to login and their username contains an uppercase letter. 

Example:
A user signs up with a username:password of "Test":"123". The user tries to go login; however, the system automatically lowercases his username to "test" and does not allow the user to login.

Before this can be fixed, I think the following question should be answered:
- Should we continue to allow usernames to be case-sensitive? "Test" and "test" are different users and Bookie currently allows this.

NSchock:
I would like to fix this issue; however, I agree we need to figure out whether usernames should be case sensitive. I personally believe they should be, but I'm not really sure I'm the one that should be making this decision, so I would like to see what everybody else thinks before fixing the issue.

mitechie:
On Tue, 18 Mar 2014, NSchock wrote:

> I would like to fix this issue; however, I agree we need to figure out
> whether usernames should be case sensitive. I personally believe they
> should be, but I'm not really sure I'm the one that should be making this
> decision, so I would like to see what everybody else thinks before fixing
> the issue.

So this was something recently changed. We decided against case sensitivity
and the code is supposed to lowercase user input. This seems like a place
got missed in that change.

Original changeset:
https://github.com/bookieio/Bookie/pull/288

If someone wants to find the case that's not lower cased, update it, and
write a test to verify it stays working that'd be great and I'm more than
happy to help guide and review the change.

## 

Rick Harding
@mitechie
http://blog.mitechie.com
http://lococast.net

HubertChen:
I hope I submitted a pull request correctly.

Before we close this issue, a few things I'm worried about are:
- Are there any conflicting usernames (e.g., "Test" and "test")? If so, who gets to keep their username?
- We need to lowercase all usernames since @sambuddhabasu last did so (~1 month ago)

Perhaps the most convenient way is to just query them in the live database and actually see how many of these issues we are dealing with. If we are dealing with an extremely low amount, then we can just fix them manually. After we get that issue solved, I believe we can use @sambuddhabasu's script to update the database.

@mitechie : What are your thoughts on the issue?

mitechie:
Thanks for the consideration. I've dealt with the Test/test cases during the initial branch to lowercase usernames. I'll check them for case again.

PR #379, opened 2014-03-25
Username is saved as lowercase during sign up

HubertChen:
See issue: #368


mitechie:
Thanks, this looks like a great change. I appreciate the good tests with the branch.

PR #380, opened 2014-04-02
Minor Bug regarding declaration of new_username 

rakeshmahadasa:


mitechie:
Thanks for this update. I've added a test in the process of merging this in. Feel free to check that out. It's often nice, even on small changes like this, to write a test that fails if that one line of code you added goes missing again. Just to be safe.

rakeshmahadasa:
Sure , thanks for the advice. will try to make it as a habit.

PR #382, opened 2014-04-04
Suggested Tags in Chrome Extension.

rakeshmahadasa:
getting error 32 while running the code.


mitechie:
What I'd guess is that you're hitting a code path that does not in fact return anything which means no response is sent to the client. The way to debug this is to use a tool called pdb to set a breakpoint and trace your way through the code to see what gets hit. 

https://docs.python.org/2/library/pdb.html

Using some tests to help check things will also help make sure you hit the areas of the code you're expecting to hit and help narrow down the issue.

rakeshmahadasa:
Issue was referenced at #392  after rebasing the code. updated my repository and this wont happen again.

Issue #390, opened 2014-04-08
footer links point to outdated GitHub repos


widox:
Both the "Support" and "Changes" links in the footer point at the old http://github.com/mitechie/Bookie/ repo.

We should update them to point at its new location here under the bookieio organization.

PR #391, opened 2014-04-08
Update bookie github links

suryatech:
Updated the links on the main site and in few other locations.
#390


mitechie:
Thanks much for the docs update.

PR #392, opened 2014-04-08
Suggest tags for chrome extension

rakeshmahadasa:


mitechie:
Landed in https://github.com/bookieio/Bookie/commit/30634bcaa5d393ba9e8c2fb7dffe53705eebc647

PR #394, opened 2014-04-08
Fixing redable content's status code issue

rakeshmahadasa:
need help in writing tests.


mitechie:
This has got a couple of mixed commits it appears.

As for the test, the trick is to make two bookmarks. You'll need the first, to generate recent_tags for the second. The second bookmark should not have a .readable. The test should then verify that the recent tags appear and there's no error as there is without this fix.

bookie/tests/test_models/test_tagmgr.py
```diff
@@ -111,3 +111,68 @@ def test_suggested_tags(self):
...
+        res = self.testapp.post('/login',
+                                params=user_data)
+        # Add a bookmark
+        res = DBSession.execute(
+            "SELECT api_key FROM users WHERE username = 'admin'").fetchone()
+        key = res['api_key']
+        url = u'http://testing_tags.com'
+        # set the readable content for the bookmark
+        path = os.getcwd()+"/bookie/tests/test_models/tag_test.txt"
+        test_content_file = open(path, 'r')
```
mitechie: you can just do

```
test_content_file = open(path, 'r').read()
```

mitechie:
Thanks, I did go in and update your test to use the factory methods available and avoid making web requests where it's not required. Take a peek as it might be useful.

rakeshmahadasa:
Ahh! I should have used factory functions to store bookmarks :\ will keep in mind next time :+1:

PR #397, opened 2014-04-09
Backend to Generate new api

rakeshmahadasa:


mitechie:
I've landed this. I did change url end point because it makes sense to use the normal http verbs for GET to find out your api key and POST to get a new one. Other than that, thanks for this branch!

PR #398, opened 2014-04-09
Api for stats information

sambuddhabasu:


bookie/views/api.py
```diff
@@ -57,6 +57,31 @@ def _api_response(request, data):
...
+    user_count = UserMgr.count()
+    pending_activations = ActivationMgr.count()
+
+    # Bookmark data and stats.
+    bookmark_count = BmarkMgr.count()
+    unique_url_count = BmarkMgr.count(distinct=True)
+    users_with_bookmarks = BmarkMgr.count(distinct_users=True)
+
+    return _api_response(request, {
+        'bookmark_data': {
```
mitechie: in thinking about this I kind of think that perhaps these should be different stats endpoints. So one for bookmark stats, one for user data stats. We can look at adding others as they're interesting.

bookie/static/js/bookie/api.js
```diff
@@ -694,6 +694,53 @@ YUI.add('bookie-api', function (Y) {
         }
     );
 
+
+    /**
+     * Fetch the user stats
```
mitechie: Fetch stats about the number of users on the site.

bookie/static/js/bookie/api.js
```diff
@@ -694,6 +694,53 @@ YUI.add('bookie-api', function (Y) {
...
+    Y.bookie.Api.route.UserStats = Y.Base.create(
+        'bookie-api-route-user-stats',
+        Y.bookie.Api.route,
+        [], {
+            initializer: function(cfg) {
+            }
+        }, {
+            ATTRS: {
+                url_element: {
+                    value: '/api/v1/userstats'
```
mitechie: api/v1/stats/users

bookie/static/js/bookie/api.js
```diff
@@ -694,6 +694,53 @@ YUI.add('bookie-api', function (Y) {
...
+    Y.bookie.Api.route.BookmarkStats = Y.Base.create(
+        'bookie-api-route-bookmark-stats',
+        Y.bookie.Api.route,
+        [], {
+            initializer: function(cfg) {
+            }
+        }, {
+            ATTRS: {
+                url_element: {
+                    value: '/api/v1/bookmarkstats'
```
mitechie: /api/v1/stats/bookmarks

bookie/routes.py
```diff
@@ -92,6 +92,14 @@ def build_routes(config):
     # NEW API
     #
 
+    # stats
+    config.add_route('api_bookmark_stats',
+                     '/api/v1/bookmarkstats',
```
mitechie: the api routes should be restful. The idea is that we're getting /api/v1/stats, what type of stats? bookmark stats. 

```
/api/v1/stats/bookmarks
```

bookie/routes.py
```diff
@@ -92,6 +92,14 @@ def build_routes(config):
...
+                     request_method='GET')
+    config.add_route('api_user_stats',
+                     '/api/v1/userstats',
```
mitechie:
```
/api/v1/stats/users
```

bookie/static/js/bookie/view.js
```diff
@@ -962,6 +962,73 @@ YUI.add('bookie-view', function (Y) {
...
+    ns.BookmarkStatsView = Y.Base.create('bookmark-stats-view', Y.View, [], {
+        /**
+         *
+         * General initializer
+         *
+         * @method initializer
+         * @param none
+         *
+         */
+        initializer: function() {
```
mitechie, line 1019: the work should be moved to the render method. The initializer method of the JS is to set things up, but not to draw it yet. That's the job of the render method.

bookie/templates/stats/dashboard.mako
```diff
@@ -1,24 +1,44 @@
...
+        <span id="bookmark_stats_unique_count"></span>
     </li>
 </ul>
+</div>
+<%def name="add_js()">
+    <script type="text/javascript">
+        // Create a new YUI instance and populate it with the required modules.
+        YUI().use('bookie-view', function (Y) {
+            Y.on('domready', function() {
+                var user_stats = new Y.bookie.UserStatsView({
```
mitechie: then this would create an instance and then we'd 

```
 user_stats.render()
```

mitechie:
This branch is good, I've got a few small suggestions. It seems it's missing some work in the dashboard view to remove the work it's doing there to build the current data in the page. If we move the work to the api we're adding some code and should be able to remove some code.

mitechie:
:testit:

mitechie:
test this please

craigmaloney:
Test pass, but it doesn't appear to work in my LXC Precise container. Please advise.

mitechie:
This branch has exposed a bug in how the Bookie JS API is constructed. What's happening is that the two views are overwriting the callbacks on each other. We end up getting the api result for the user data, and that ends up getting fed into the bookmark success callback. 

This bug is not with this branch, but due to the fact that there's a long standing bug this is exposing. I'm trying to chase it down, but it's in some deep JS magic code.

mitechie:
This got some updates and is landed.

sambuddhabasu:
@mitechie Will try looking into the bug. Thanks for the update.

Issue #387, opened 2014-04-07
Missing 'http://' in the URL while adding a new bookmark


sambuddhabasu:
Previously, while adding a new bookmark, a 'http://' used to get added in the beginning of the Url, which does not happen anymore.
This can be seen at https://bmark.us/{username}/new

sambuddhabasu:
The same error seems to occur for the local instance too.

PR #399, opened 2014-04-09
Update view.js

preddy5:
there is a bug in 'on' function in YUI so using 'set' function


preddy5:
issue #387

bookie/static/js/bookie/view.js
```diff
@@ -954,9 +954,9 @@ YUI.add('bookie-view', function (Y) {
         initializer: function () {
             // The url is not editable. So if we don't have the input, assume
             // it's an edit and skip.
-            var url_input = Y.one('input#url');
+            var url_input = Y.one('#url');
             if (url_input) {
-                url_input.on('onblur', this._check_and_edit_url);
```
mitechie, line 959: I think the bug here is that it's not 'onblur' but just 'blur'. Did you test with that? You might also need to make sure to require the event-focus YUI module. 

http://yuilibrary.com/yui/docs/event/focus.html

so that would need to be in the YUI.use() block.

mitechie:
I've landed a fix for this issue. You can see what was involved over at:

https://github.com/bookieio/Bookie/commit/4c0779ac4a499f6feeea3f274fa6c62556dd9ebf

PR #400, opened 2014-04-09
Allow user to generate new API key from his accounts page

suryatech:


mitechie:
Tweaked and landed this to go with the other branch that implemented the back end code. Thanks for the update!

PR #404, opened 2014-04-14
Move css before first_bookmark.

craigmaloney:
If first_bookmark has issues then the CSS will not build. Changing the order so this shouldn't be as much of an issue.


mitechie:
Thanks for the update!

PR #416, opened 2014-04-16
Setup Vagrant for easy non-Ubuntu dev env.

kadams54:
Per the updated `README.rst`, you can now get Bookie running on non-Ubuntu machines with:

```
$ vagrant up
$ vagrant ssh
% cd /vagrant && make run
```

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env sh
+
+apt-get update
+apt-get install -y make
+
+cd /vagrant
+export NONINTERACTIVE=1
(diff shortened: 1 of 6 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+apt-get update
+apt-get install -y make
+
+cd /vagrant
+export NONINTERACTIVE=1
+make sysdeps && make css
```
mitechie: should this also make js?

mitechie:
Thanks!

PR #422, opened 2014-04-17
Fix incorrect example in api docs

suryatech:


mitechie:
Thanks for this doc update.

PR #426, opened 2014-04-18
Remove last_bmark from bmark_get api call

rakeshmahadasa:
Removed last_bmark in response also as we are not using it now!! are we going to purge the last_bmark completely from the code base?? 

test will be updated once the code was approved!


mitechie:
Landed this with a couple of small tweaks removing some more model code and fixing tests from trunk.

Issue #385, opened 2014-04-07
allow for disabling new signups


mitechie:
If hosting your own, you might want to disable others from signing up on your hosted instance. This should be a documented .ini setting.

PR #427, opened 2014-04-18
adding ability to make signup private process

preddy5:
#385


bookie/templates/index.mako
```diff
@@ -4,16 +4,20 @@
...
-            <form id="#signup_form" action="signup_process" method="POST">
-                <ul>
-                    <li>
-                        <input type="email" id="email" name="email"
-                        placeholder="email address" />
-                        <input type="submit" id="send_signup" name="send_signup" value="Sign Up" />
-                    </li>
-                </ul>
-            </form>
+            % if signup:
```
mitechie, line 7: how about making this more explicit. 'signup_enabled'

bookie/tests/test_auth/test_signup.py
```diff
@@ -204,3 +206,42 @@ def testSignupWorks(self):
             email,
             act.user.email,
             "The activation email is the correct one.")
+
+
+class TestPrivateHosting(TestCase):
```
mitechie, line 211: You can extend the WebTest from the test/**init**.py and have the setup waiting for you.I know you need to tweak the config, but perhaps we can do that by sticking an override_config attribute on the class?

bookie/views/__init__.py
```diff
@@ -30,7 +31,12 @@ def home(request):
         username = username.lower()
 
     if not request.user:
-        return {}
+        if asbool(request.registry.settings.get('signup_process')):
```
mitechie, line 34: so then I'd make this match. 'signup_enabled'. It makes it clear what true/false do to this.

test.ini
```diff
@@ -21,6 +21,9 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_test
 
+#Change to false to disable signups
+signup_process = true
```
mitechie, line 25: this also needs to go into the test_pgsql.ini. We've recently set that up on the CI server. http://ci.bookie.io:8080

mitechie:
This looks really great. Thanks for the awesome branch. I've got a couple of small suggestions and will try it out later.

bookie/tests/test_auth/test_signup.py
preddy5, line 211: awesome idea, i'll work on it

mitechie:
I'm holding off on landing this as you expressed interest in updating the TestCase code.

PR #428, opened 2014-04-19
Suggested Tags when using Bookmarklet

rakeshmahadasa:
Need help in updating tests.
Made some major changes.
1.New Bookmarks wont end up at tag suggestions
2. No more recent tags on website 
3. When user uses Bookmarklet , tags will be suggested as the same way we are doing it in chrome extension
4. Tags based on url and description are added by default.


mitechie:
Let's make sure to rebase develop into this and update it with the latest changes. This assumes you've got the shortcuts from the HACKING doc supplied to setup the remote and such. 

```
git checkout develop
git sync-bookie
git checkout tag_improve
git rebase -i develop
```

rakeshmahadasa:
Code was upto date when pulled , will rebase it with the current changes now :+1:  what about the code?? if its ok I will work on updating tests.

mitechie:
Landed in pr 440

PR #440, opened 2014-04-27
Improve tag suggestions for web ui users.

mitechie:
- Make sure that you get the title/url suggestions as well as the content
  suggestions when saving a bookmark via the webui.

Landing on behalf of infinitum. Will try this out to test the ci automated lander.


mitechie:
:shipit:

Issue #446, opened 2014-05-01
search_content parameter doesn't appear to be working


dyladan:
using the api endpoint `/:username/bmarks/search/:terms` or `/bmarks/search/:terms`, the parameter `search_content=true` doesn't appear to affect the search results in any way. I have tried this with 20 or so different searches.

dyladan:
I think I may have found the problem. https://github.com/bookieio/Bookie/blob/develop/bookie/views/api.py#L502 and http://docs.bmark.us/en/latest/api/user.html#username-bmarks-search-terms seem to be at odds with each other. The documentation specifies `search_content` and the code comment specifies `with_content`. FWIW the `with_content` parameter does affect the search results.

dyladan:
https://github.com/bookieio/Bookie/blob/develop/bookie/views/api.py#L527 makes it clear that the documentation should say `with_content` instead of `search_content`. Creating a pull request presently.

PR #447, opened 2014-05-01
fixed search API endpoint documentation

dyladan:
Used to specify the parameter search_content when it should
have specified with_content

fixes/invalidates issue #446


mitechie:
:shipit: thanks for the update

Issue #445, opened 2014-05-01
with_content parameter doesn't work


dyladan:
on the api endpoint `/:username/bmarks` the `with_content=true` parameter doesn't appear to be doing anything as shown by the fact that the following two URLs return the exact same JSON.

https://bmark.us/api/v1/dyladan/bmarks?count=1&with_content=true
https://bmark.us/api/v1/dyladan/bmarks?count=1

dyladan:
This may affect other endpoints as I have only tested this one. It is worth noting that all of the following routes use the same function in api.py
- `api_bmarks`
- `api_bmarks_user`
- `api_bmarks_tags`
- `api_bmarks_user_tags`

edit: tested all of the above endpoints and none of them responds to `with_content`

dyladan:
full text is returned properly when the bmark is called using `/:username/bmark/:hash_id` endpoint

dyladan:
if you change this line https://github.com/bookieio/Bookie/blob/develop/bookie/views/api.py#L412 to be `if 'with_content' in params and params['with_content'] != 'false':` it appears to be working on my local copy at least. I have a working branch but would like help writing the test for this.

mitechie:
This now functions properly.

PR #448, opened 2014-05-01
fixed api with_content for :username/bmarks and /bmarks

dyladan:
wrote test to ensure this does not regress


dyladan:
This fixes #445

bookie/views/api.py
```diff
@@ -409,7 +409,7 @@ def bmark_recent(request, with_content=False):
         return_obj['url'] = res.hashed.url
         return_obj['total_clicks'] = res.hashed.clicks
 
-        if with_content:
+        if 'with_content' in params and params['with_content'] != 'false':
```
mitechie: there's an asbool thing that will help check for false, False, FALSE, etc. Check out http://www.nullege.com/codes/search/pyramid.settings.asbool

I'd prefer we use that to verify that it's true and false. We should have a test for both cases. 

Thanks for the work on this and investigation.

mitechie:
Comment in the diff. Let me know if you need a hang with anything.

mitechie:
Testing out if :sos: will force a test build on another person's branch.

dyladan:
@mitechie switched to use asbool and added a false test

mitechie:
Cool, can you rebase into a single commit and then I'll have it run through CI

dyladan:
rebased and pushed

mitechie:
Hopefully CI is now working with update. :sos:

mitechie:
ok to test

mitechie:
ok to test

Issue #449, opened 2014-05-02
/:username/bmarks/search/:terms appears to be broken


dyladan:
`https://bmark.us/api/v1/dyladan/bmarks/search/coroutines` returns no bookmarks while `https://bmark.us/api/v1/bmarks/search/coroutines` does return bookmarks, at least one of which is `username=dyladan`

dyladan:
ignore this it's working fine

dyladan:
reopening this since i figured out what was causing my original issue.

`https://bmark.us/api/v1/dyladan/bmarks/search/coroutines&with_content=true` returns no results
`https://bmark.us/api/v1/dyladan/bmarks/search/coroutines` does return results

`with_content=true` works fine on `/bmarks/search/:terms` endpoint but on the `/:username/bmarks/search/:terms` endpoint it appears that it thinks `&with_content=true` is a part of the search term and not a parameter.

here is the full json response for the non-working endpoint

```
{"username": "dyladan", "search_results": [], "with_content": false, "phrase": "coroutines&with_content=true", "page": 0, "result_count": 0}
```

dyladan:
so apparently none of the params are working. they are all just a part of the search string. I found this issue and it's pretty simple fix. writing a test now

dyladan:
do not merge this. I was misunderstanding the problem

dyladan:
https://bmark.us/api/v1/dyladan/bmarks/search/coroutines
returns 1

https://bmark.us/api/v1/dyladan/bmarks/search/coroutines?with_content=true
returns 0 results

https://bmark.us/api/v1/bmarks/search/coroutines
returns 5 results

https://bmark.us/api/v1/bmarks/search/coroutines?with_content=true
returns 10 results

In none of these queries is the above problem where content=true was showing up in the phrase. I think that was just because i was using & instead of ?

The non-username search appears to be working as expected.

The user specific search appears to work if there are no params, but if you add `with_content=true` or `count=` params no results are returned

dyladan:
this appears related to #450

mitechie:
Closing as lost in history and there's something else up elsewhere.

PR #451, opened 2014-05-02
fixed 449 and added a test

dyladan:
also, removes the GET parameters from the search phrase by splitting on the first '?'


bookie/views/api.py
```diff
@@ -504,10 +504,10 @@ def search_results(request):
 
     """
     mdict = request.matchdict
-    rdict = request.GET
+    rdict = request.params
 
     if 'terms' in mdict:
-        phrase = " ".join(mdict['terms'])
+        phrase = " ".join(mdict['terms']).split("?")[0]
```
mitechie: this seems broken or  a bug. Pyramid should not include the query string in the matchdict. I'll have to test this out and see about working upstream. We upgraded to pyramid 1.5 recently and I wonder if there's a known bug or update to webob or pyramid we should be getting.

bookie/views/api.py
```diff
@@ -504,10 +504,10 @@ def search_results(request):
 
     """
     mdict = request.matchdict
-    rdict = request.GET
+    rdict = request.params
```
mitechie: why the change here?

dyladan: did this as part of trying to pinpoint the issue. I'm pushing a commit now that switches it back

dyladan:
amended the commit to switch back to using GET

bookie/views/api.py
dyladan: not sure. i just know the GET parameters were being included in the terms and this is a working fix for that. If there is a better or more robust solution for this i'd be more than willing to do that instead

dyladan:
see #449 closing

dyladan:
this apparently does still fix the issue and the tests still pass. reopening

mitechie:
Cool, can you rebase this back down to one commit and I'll test/qa this over the weekend. Thanks for looking into it.

dyladan:
i was wrong. closing. The issue is still open, i'm just not sure what causes it.

PR #456, opened 2014-05-21
Updated alembic version

sambuddhabasu:
A newer version of alembic(0.6.5) is out and the requirements.txt need to be updated accordingly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-alembic==0.3.4
+alembic==0.6.5
 amqp==1.0.12
 anyjson==0.3.3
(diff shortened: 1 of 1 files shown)


mitechie:
Please make sure to provide a description in all pull requests to aid code reviewers. 

What change is this
What is it meant to do
Is it related to any other items
Are there precise steps to be able to QA the change

mitechie:
This cannot land and won't pass CI until the download-cache is updated. I'll wait on confirmation that it's updated before reviewing.

mitechie:
:sos:

PR #457, opened 2014-05-21
Updated alembic version

sambuddhabasu:
A newer version of alembic(0.6.5) is out and the requirements.txt need to be updated accordingly.


mitechie:
this breaks the build and needs the migration fixes to land. Closing for now. I look forward to the updated branch.

PR #458, opened 2014-05-21
All exceptions are now handled in database migration

sambuddhabasu:
A different kind of exception was being raised, which caused the errors. Any kind of exceptions are now handled.
Also, alembic version has been updated to 0.6.5

### DIFF
--- a/dbversions/versions/44dccb7b8b82_update_username_to_l.py
+++ b/dbversions/versions/44dccb7b8b82_update_username_to_l.py
@@ -24,7 +24,7 @@ def upgrade():
     try:
         op.drop_constraint("bmarks_username_fkey", "bmarks")
         print 'dropped constraint'
-    except sa.exc.OperationalError:
+    except (sa.exc.OperationalError, NotImplementedError) as exc:
         # If it's not supported then pass
         pass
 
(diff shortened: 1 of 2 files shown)


dbversions/versions/44dccb7b8b82_update_username_to_l.py
```diff
@@ -24,7 +24,7 @@ def upgrade():
     try:
         op.drop_constraint("bmarks_username_fkey", "bmarks")
         print 'dropped constraint'
-    except sa.exc.OperationalError:
+    except Exception as e:
```
mitechie: this needs to be more specific to catch the right kind of exception. We're expecting sqlite to throw an error, but that's specific. If there's a syntax error in the migration, or something else, we don't want that caught and ignored here.

mitechie:
Thanks for the pull request. I'd expect this to also have the version update in the requirements.txt file? 

I've added a comment, this isn't quite ready to land yet, but I think you're on the right path updating the exception handling.

mitechie:
Thanks for the update. The code looks good to me. :+1: 

Once @pradyu1993 reviews it and checks it out this is ok to land via the shipit command.

preddy5:
+1

mitechie:
:shipit:

PR #461, opened 2014-05-26
Database migration to support private bookmarks

sambuddhabasu:
Exisitng bookmarks should be public

### DIFF
--- a/bookie/tests/test_models/test_privatebmark.py
+++ b/bookie/tests/test_models/test_privatebmark.py
@@ -1,0 +1,67 @@
+"""Test private bookmark support"""
+
+from pyramid import testing
+
+from bookie.models import Bmark
+from bookie.models.auth import User
+
+from bookie.tests import empty_db
+from bookie.tests import gen_random_word
+from bookie.tests import TestDBBase
+
+
+class TestPrivateBmark(TestDBBase):
+    """Handle private bookmark checks"""
+
+    def setUp(self):
+        """Setup Tests"""
+        from pyramid.paster import get_app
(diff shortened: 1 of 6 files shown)


mitechie:
Have you tested the downgrade step? 

I'd like to see one test added to the models test to verify the field is there and that the default is set properly. 

I think this will be great with that test and verification that the downgrade works manually. 

Thanks for the work. I'll add the +1 once the pull request is updated.

craigmaloney:
Note: I tried this with sqlite3 and it appears that even after doing the upgrade `.schema bmarks` didn't show the column, even though the downgrade and subsequent upgrade steps told me it was still there. So sqlite3 may be more tricky than we realize.

bookie/tests/test_models/test_privatebmark.py
```diff
@@ -0,0 +1,68 @@
...
+        bmark = Bmark(
+            url=gen_random_word(12),
+            username=user.username
+        )
+        self.assertEqual(
+            True,
+            bmark.is_private,
+            'Default value of is_private should be True')
+
+    def test_is_private_value_passed(self):
```
mitechie: we might as well also have a check that if the argument is passed as False it ends up being public. This helps make sure we don't have a hard coded value we can't unset.

mitechie:
Thanks! Looks good with the note on one more test that should probably be added. After that, it's good to land from me. :+1:

preddy5:
:+1:

sambuddhabasu:
:shipit:

PR #463, opened 2014-05-28
Private count tests added for models

sambuddhabasu:
Models tests have been updated to take into account the private flag with private and non-private count tests.

### DIFF
--- a/bookie/tests/test_models/test_bmarkmgr.py
+++ b/bookie/tests/test_models/test_bmarkmgr.py
@@ -42,13 +42,33 @@ class TestBmarkMgrStats(TestDBBase):
         for i in range(ct):
             b = Bmark(
                 url=gen_random_word(12),
-                username=user.username
+                username=user.username,
+                is_private=False
             )
             b.hash_id = gen_random_word(3)
             DBSession.add(b)
 
         ct = BmarkMgr.count()
-        self.assertEqual(5, ct, 'We should have a total of 5: ' + str(ct))
+        self.assertEqual(5, ct,
+                         'We should have 5 public bookmarks: ' + str(ct))
+
+    def test_total_ct_accounts_for_privacy(self):
+        """Verify that our total count method is working"""
+        ct = 5
+        user = User()
+        user.username = gen_random_word(10)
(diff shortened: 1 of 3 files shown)


bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -123,6 +123,44 @@ def test_per_user(self):
         ct = BmarkMgr.count(username=usercommon.username)
         self.assertEqual(2, ct, 'We should have a total of 2: ' + str(ct))
 
+    def test_per_user_is_private(self):
```
mitechie: test_per_user_accounts_for_privacy

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -123,6 +123,44 @@ def test_per_user(self):
...
+        d = Bmark(
+            url=gen_random_word(10),
+            username=usercommon.username,
+            is_private=False,
+        )
+        DBSession.add(d)
+        DBSession.flush()
+
+        ct = BmarkMgr.count(username=usercommon.username, is_private=False)
+        self.assertEqual(2, ct, 'We should have a total of 2: ' + str(ct))
```
mitechie: "We should only have 2 private bookmarks"

sambuddhabasu: Shouldn't the string be "We should only have 2 public bookmarks" ?

mitechie: oh yea, it'd help if I read things properly

preddy5:
:+1:

mitechie:
This branch looks good. We should file a bug about the pgsql failures to investigate. It seems it started to get flaky after the migration to add is_private. Can you file that bug and add a card for that bug to trello we can look at soon?

Thanks for the change.

sambuddhabasu:
So, do you think that this branch is good to be merged?
Also, I will file a bug shortly and also update the trello board.

mitechie:
Sorry yes to land. :+1:

sambuddhabasu:
:shipit:

PR #464, opened 2014-05-28
First bookmark should be public

sambuddhabasu:


mitechie:
:shipit: thanks for the update

PR #466, opened 2014-06-01
Adding BaseConnection and TwitterConnection db migration and model

preddy5:


sambuddhabasu:
I have reviewed it :+1:

bookie/models/auth.py
```diff
@@ -269,6 +269,8 @@ class User(Base):
     api_key = Column(Unicode(12))
     invite_ct = Column(Integer, default=0)
     invited_by = Column('invited_by', Unicode(255))
+    BaseConnection = relation(BaseConnection,
```
mitechie, line 272: you can move this relation to the connection side I believe.

mitechie, line 272: hmm, never mind this is cool as it is.

bookie/models/social.py
```diff
@@ -0,0 +1,43 @@
+
+from sqlalchemy import Column
+from sqlalchemy import DateTime
+from sqlalchemy import ForeignKey
+from sqlalchemy import Integer
+from sqlalchemy import UnicodeText, Unicode, Boolean
```
mitechie: please group imports like this

```
from sqlalchemy import (
    Column,
    DateTime,
    ...
    UnicodeText,
    Unicode,
    Boolean)
```

but make sure they're listed alphabetically.

bookie/models/social.py
```diff
@@ -0,0 +1,43 @@
...
+from bookie.models import Base
+
+
+class BaseConnection(Base):
+    """Table to store User basic social information"""
+    __tablename__ = "BaseConnection"
+    id = Column(Integer, primary_key=True)
+    username = Column(Unicode(255),
+                      ForeignKey('users.username'))
+    type = Column(UnicodeText())
```
mitechie: any reason for UnicodeText vs just Unicode?

bookie/models/social.py
```diff
@@ -0,0 +1,43 @@
...
+        'polymorphic_identity': 'BaseConnection',
+        'polymorphic_on': type
+    }
+
+
+class TwitterConnection(Base):
+    """ Table to store User Twitter information"""
+    __tablename__ = "TwitterConnection"
+
+    user_id = Column(Integer, ForeignKey('BaseConnection.id'),
```
mitechie: this isn't user_id but the connection_id correct?

bookie/models/social.py
```diff
@@ -0,0 +1,43 @@
...
+                     primary_key=True)
+
+    # User Twitter Data
+    twitter_uid = Column(UnicodeText())
```
mitechie: I think all these unicodetext can just be unicode. The UnicodeText would be for large bodies of text, like the html page content, but not for shorter more regular bits of text like usernames or access keys.

bookie/models/social.py
```diff
@@ -0,0 +1,43 @@
...
+    twitter_access_key = Column(UnicodeText())
```
mitechie: since this is a TwitterConnection I think we can remove the twitter_ prefix to everything. Just access_key, uid, etc.

bookie/models/auth.py
```diff
@@ -269,6 +269,8 @@ class User(Base):
...
+                              uselist=False, backref="users")
```
mitechie: but actually this should be a uselist. A user might have many social network connections in the future. uselist forces a 1->1 relation, but we want a one user -> many connections relationship.

mitechie:
Thanks for this update Prady, it's close but I think there's a few things to correct before landing. Let me know if you have any questions.

dbversions/versions/352bb5f4fff9_adding_baseconnection_and_.py
```diff
@@ -0,0 +1,45 @@
...
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('BaseConnection',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('username', sa.Unicode(length=255), nullable=True),
+    sa.Column('type', sa.Unicode(), nullable=True),
+    sa.Column('is_active', sa.Boolean(), nullable=True),
+    sa.Column('last_connection', sa.DateTime(), nullable=True),
+    sa.ForeignKeyConstraint(['username'], ['users.username'], ),
```
mitechie: in working with the mysql tests I've noticed it's best if we name the constraints. so this would be baseconnection_username_fkey

dbversions/versions/352bb5f4fff9_adding_baseconnection_and_.py
```diff
@@ -0,0 +1,45 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('BaseConnection',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('username', sa.Unicode(length=255), nullable=True),
```
mitechie: I think you can just Unicode(255) without the length=

It should probably be set like that for the other Unicode() fields as well.

dbversions/versions/352bb5f4fff9_adding_baseconnection_and_.py
mitechie: An example of what I mean is here: https://github.com/bookieio/Bookie/pull/470/files#diff-810636c25804a806e0bcc7a670932f56R68

mitechie:
This looks almost ready. As part of the mysql work I've noticed that we should be more specific when creating constraints (like the foreign key ones) and the Unicode() references should be Unicode(255).

After that and a round of QA I think this will be ready to move forward and land.

mitechie:
:+1: thanks for the updates

preddy5:
:shipit:

PR #467, opened 2014-06-03
Fix to the pgsql tests

sambuddhabasu:


sambuddhabasu:
The pgsql tests are running fine now.

PR #468, opened 2014-06-04
By default, bookmarks stored should be private

sambuddhabasu:
By default, the store function of BmarkMgr should be storing private bookmarks unless specified otherwise.

### DIFF
--- a/bookie/tests/test_models/test_privatebmark.py
+++ b/bookie/tests/test_models/test_privatebmark.py
@@ -2,7 +2,10 @@
 
 from pyramid import testing
 
-from bookie.models import Bmark
+from bookie.models import (
+    Bmark,
+    BmarkMgr,
+)
 from bookie.models.auth import User
 
 from bookie.tests import empty_db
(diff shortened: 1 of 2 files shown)


mitechie:
This should have tests around store() and updating current code to specify the use. I know store is used in the import code specifically.

bookie/tests/test_models/test_privatebmark.py
```diff
@@ -65,3 +68,42 @@ def test_is_private_false(self):
         self.assertEqual(
             False,
             bmark.is_private)
+
+    def test_storing_private_true(self):
+        """Verify the value of is_private is True"""
```
mitechie: "Verify the value of is_private defaults to true"

mitechie:
Thanks for adding the test. :+1: for this but I want to make sure we keep in mind that any imports that go on using this code will end up with all imported bookmarks as private.

preddy5:
:+1:

sambuddhabasu:
:shipit:

PR #469, opened 2014-06-05
Update dependencies.

mitechie:
- Remove psycopg from the base install as it defaults to sqlite.
- Move the pyinotify to the dev-requirements so that OSX base install can
  work.


mitechie:
:shipit:

PR #471, opened 2014-06-06
Tests added for api calls to support private bookmarks

sambuddhabasu:


bookie/tests/test_api/test_base_api.py
```diff
@@ -149,6 +154,42 @@ def test_add_bookmark(self):
         self.assertTrue(
             'description": "Bookie"' in res.body,
             "Should have Bookie in description: " + res.body)
+        self.assertTrue(
+            '"is_private": false' in res.body,
+            "Should have is_private as false: " + res.body)
+        self._check_cors_headers(res)
+
+    def test_add_bookmark_accounts_for_privacy(self):
```
mitechie: I'd just go simpler here. "test_add_private_bookmark"

bookie/tests/test_api/test_base_api.py
```diff
@@ -285,22 +326,64 @@ def test_bookmark_fetch_fail(self):
...
         self._get_good_request()
 
-        # test that we get a 404
+        # test that we get a 200
         res = self.testapp.get(
             '/api/v1/admin/bmark/{0}'.format(GOOGLE_HASH),
             status=200)
         self._check_cors_headers(res)
 
+    def test_bookmark_diff_user_accounts_for_privacy(self):
```
mitechie: test_no_access_to_private_bookmark

bookie/views/api.py
```diff
@@ -153,6 +153,14 @@ def bmark_get(request):
     # The hash id will always be there or the route won't match.
     bookmark = BmarkMgr.get_by_hash(hash_id, username=username)
 
+    if bookmark:
+        if bookmark.is_private:
+            if request.user:
```
mitechie: I think this part should be pulled into a function as I think we'll need to use it more than once. 

```
if request.user and not bookmark.has_access(request.user.username):
    bookmark = None
```

So this would require adding a has_access method to the Bmark class that would accept a username and check it against the current user. It might as well also do the private check as well. 

Then the code would read just as

```
if bookmark and not bookmark.has_access(request.user.username):
    bookmark = None
```

mitechie:
This seems like a really good start Sam. I'm having some issues with the names of the test functions, but they're tough as they're complex. 

I've got one suggest for making the access checks a bit easier. Let me know if that makes sense or not.

sambuddhabasu:
Thanks Rick, I think the idea of having the has_access method looks good. It would make the code cleaner and more understandable.

bookie/models/__init__.py
```diff
@@ -683,6 +683,10 @@ def update_tags(self, tag_string):
         """Given a tag string, split and update our tags to be these"""
         self.tags = TagMgr.from_string(tag_string)
 
+    def has_access(self, username):
+        """Check if the passed username is same as that of the bookmark"""
+        return self.username == username
```
mitechie: this needs to check if the bookmark is private.

bookie/models/__init__.py
```diff
@@ -683,6 +683,21 @@ def update_tags(self, tag_string):
         """Given a tag string, split and update our tags to be these"""
         self.tags = TagMgr.from_string(tag_string)
 
+    def has_access(self, user):
+        """Check if a user has access to view a bookmark"""
+        if self.is_private:
+            if user:
```
mitechie: user should not be passed in. The bmark model should not know how to parse out the attributes of the user object. It should just take the string.

bookie/tests/test_models/test_bmark.py
```diff
@@ -0,0 +1,99 @@
...
+            is_private=True,
+        )
+        b.hash_id = gen_random_word(3)
+        DBSession.add(b)
+        res = b.has_access(user)
+        self.assertEqual(False, res)
+        # Also check if user is None.
+        user = None
+        res = b.has_access(user)
+        self.assertEqual(False, res)
```
mitechie, line 99: These tests are great covering the possible cases.

bookie/views/api.py
```diff
@@ -153,6 +153,9 @@ def bmark_get(request):
     # The hash id will always be there or the route won't match.
     bookmark = BmarkMgr.get_by_hash(hash_id, username=username)
 
+    if bookmark and not bookmark.has_access(request.user):
```
mitechie: <3 how this line reads now.

mitechie:
:+1: and can land this after changing to not pass the whole User object into the has_access function. It's not a good idea to pass too much knowledge about domain models and such into simple functions such as well.

preddy5:
:+1:

mitechie:
:+1: thanks for the hard work on this branch. Hopefully these changes will make it nice and easy to move forward.

sambuddhabasu:
:shipit: Yes, working on updating the api code completely.

PR #472, opened 2014-06-10
api end point to retrieve list of connections

preddy5:


bookie/tests/factory.py
```diff
@@ -64,6 +65,20 @@ def make_tag(name=None):
     return Tag(name)
 
 
+def make_base_connection():
```
mitechie: this is making a twitter connection. I think we should call it what it is please.

bookie/tests/test_api/test_social_api.py
```diff
@@ -0,0 +1,37 @@
+
+import json
+import transaction
+from unittest import TestCase
+from pyramid import testing
+
+from bookie.tests import factory
+
+
+class TestSocialConnectionsApi(TestCase):
```
mitechie: This can extend the WebView base class and the sestup/teardown should be ok automatically. 

We do need to make sure that the teardown helper empty_db() deletes all records from the connection tables.

bookie/tests/test_api/test_social_api.py
```diff
@@ -0,0 +1,37 @@
...
+        testing.tearDown()
+
+    def testBaseApi(self):
+        factory.make_base_connection()
+        # Login a user and then test the validation of social api
+        user_data = {'login': u'admin',
+                     'password': u'admin',
+                     'form.submitted': u'true'}
+        transaction.commit()
+        self.testapp.post('/login', params=user_data)
```
mitechie: we don't need to login in an api request. We just need to pass the proper url credentials of the username and api_key fields. Check the other api tests for examples.

bookie/views/api.py
```diff
@@ -943,6 +943,28 @@ def invite_user(request):
...
+        social_info['username'] = connection.username
+        if connection.type == 'TwitterConnection':
+            social_info[connection.type] = {}
+            social_info[connection.type]['twitter_username'] = \
+                connection.twitter_username
+            social_info[connection.type]['refresh_date'] = \
+                str(connection.refresh_date)
+        social_info['last_connection'] = str(connection.last_connection)
+        res.append(social_info)
+    return _api_response(request, {'Social_Connections': res})
```
mitechie: no capital letters please. This should work like the other api calls that return sets of data. 

For instance: http://docs.bmark.us/en/latest/api/user.html#id12

We should return a count attribute in the response as well. 

I think this could just be 'result'. I guess what do the other api calls that return data sets use?

bookie/views/api.py
```diff
@@ -943,6 +943,28 @@ def invite_user(request):
...
+def social_connections(request):
+    rdict = request.matchdict
+    username = rdict.get('username', None)
+    connections = BaseConnection.query.filter(
+        BaseConnection.username == username)
+    res = []
+    for connection in connections:
+        social_info = {}
+        social_info['username'] = connection.username
+        if connection.type == 'TwitterConnection':
```
mitechie: We should make this a constant then. Matching strings is a pain and lint won't catch typos in them. I'd suggest that a constant be created in the models. 

```
TwitterConnection = 'TwitterConnection'
```

Then you can use that in place of the string in the:
     **mapper_args** = {
        'polymorphic_identity': 'TwitterConnection'
    }

And here you can import that and use it to compare the type against.

mitechie:
Thanks for this branch. This is a great start. I've made a few suggestions to make sure the tests work cleanly and clean up the db properly. 

The API endpoint also needs to be documented as the other endpoints are as part of this branch.

bookie/tests/test_api/test_social_api.py
```diff
@@ -0,0 +1,31 @@
...
+class TestSocialConnectionsApi(TestViewBase):
+    """Check if SocialConnections API returns all the added accounts"""
+
+    def testBaseApi(self):
+        factory.make_twitter_connection()
+        transaction.commit()
+
+        res = self.app.get('/api/v1/admin/social_connections',
+                           status=403)
+        self.assertEqual(
```
mitechie: there should be two tests. One that verifies it's forbidden, and one successful. Try not to make multiple assertions in one tests to help aid in making it clear what should work and so that if you're debugging a failing test you can easily run the one in isolation.

bookie/views/api.py
```diff
@@ -35,7 +35,7 @@
 from bookie.models.auth import UserMgr
 from bookie.models.stats import StatBookmarkMgr
 from bookie.models.queue import ImportQueueMgr
-
+from bookie.models.social import BaseConnection, twitter_connection
```
mitechie:
```
from bookie.models.social import (
    BaseConnection,
    twitter_connection
)
```

bookie/views/api.py
```diff
@@ -943,6 +943,29 @@ def invite_user(request):
         })
 
 
+@view_config(route_name="api_social_connections", renderer="jsonp")
+@api_auth('api_key', UserMgr.get)
+def social_connections(request):
+    rdict = request.matchdict
+    username = rdict.get('username', None)
+    connections = BaseConnection.query.filter(
```
mitechie: Views should never make queries. That's why we've got the model Mgr classes. We should have a SocialMgr or something that performs the query based on arguments passed into it from the View. The manager then returns the data just via casting to a dict. See things like https://github.com/bookieio/Bookie/blob/develop/bookie/views/api.py#L135 for an example of using a manager to get data and then building the response.

mitechie:
Thanks for the updates. The docs look good and we've talked in irc about adding the model tests. I also think we need the move to the model manager before this can land. Let me know if you have any questions or need a hand. Hopefully the existing examples help provide some good guidance.

bookie/views/api.py
```diff
@@ -943,6 +946,28 @@ def invite_user(request):
         })
 
 
+@view_config(route_name="api_social_connections", renderer="jsonp")
+@api_auth('api_key', UserMgr.get)
+def social_connections(request):
+    rdict = request.matchdict
+    username = rdict.get('username', None)
+    connections = SocialMgr.get_all_connections(username)
```
mitechie: so if this returns a list of conntections, you should be able to do:

```
res = [dict(con) for con in SocialMgr.get_all_connections(username)]
return _api_response(
    request, {
        'count': len(res),
        'social_connections': res
}) 
```

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+
+Example
+'''''''
+::
+
+    requests.get('http://127.0.0.1:6543/api/v1/admin/social_connections/api_key=12345..')
+    >>> {
+            "count": 1
+            "social_connections": [
+              {"username": "admin",
```
mitechie: move the { up to the end of the previous line please

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+::
+
+    requests.get('http://127.0.0.1:6543/api/v1/admin/social_connections/api_key=12345..')
+    >>> {
+            "count": 1
+            "social_connections": [
+              {"username": "admin", 
+               "last_connection": "2014-06-12 17:39:41.855184",
+               "TwitterConnection": {
+                                     "twitter_username": "bookie",
```
mitechie: and move this back so that it's just 4 spaces in from the parent line

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+                                     "refresh_date": "2014-06-12 17:39:41.855202"
+                                    }
+              }
```
mitechie, line 586: this should be }, { I think

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+              {"username": "admin", 
+               "last_connection": "2014-06-12 17:41:09.720954",
+               "TwitterConnection": {
+                                     "twitter_username": "bookie",
```
mitechie: and another move back so it's just 4 spaces in from the previous line.

bookie/views/api.py
preddy5: @mitechie this will return access_key and access_secret,

sambuddhabasu:
I think after making the changes that @mitechie has mentioned, it is good :+1:

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+
+Example
+'''''''
+::
+
+    requests.get('http://127.0.0.1:6543/api/v1/admin/social_connections/api_key=12345..')
+    >>> {
+            "count": 2
+            "social_connections": [
+            {"username": "admin",
```
mitechie: Please make the open { on the line before

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+            "count": 2
+            "social_connections": [
+            {"username": "admin", 
+             "last_connection": "2014-06-12 17:39:41.855184",
+             "TwitterConnection": {
+                "twitter_username": "bookie",
+                "refresh_date": "2014-06-12 17:39:41.855202"
+              }
+            },
+            {"username": "admin",
```
mitechie: please move the open { up to the line before

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+            {"username": "admin", 
+             "last_connection": "2014-06-12 17:39:41.855184",
+             "TwitterConnection": {
+                "twitter_username": "bookie",
+                "refresh_date": "2014-06-12 17:39:41.855202"
+              }
+            },
+            {"username": "admin", 
+             "last_connection": "2014-06-12 17:41:09.720954",
+             "TwitterConnection": {
```
mitechie: I think this is lower cased looking at the code you've got here.

docs/api/user.rst
```diff
@@ -541,6 +541,50 @@ Example
...
+                  "twitter_username": "bookie",
+                  "refresh_date": "2014-06-12 17:41:09.720954"
+              }
```
mitechie, line 586: We also need to export a type attribute that states this is a twitter connection.

mitechie:
:+1: thanks for all the hard work on this branch!

preddy5:
:shipit:

mitechie:
:shipit: thanks

preddy5:
:shipit:

PR #473, opened 2014-06-11
Updated models and corresponding tests

sambuddhabasu:


bookie/models/__init__.py
```diff
@@ -401,6 +401,9 @@ def find(limit=50, order_by=None, page=0, tags=None, username=None,
 
         offset = limit * page
 
+        if requested_by != username:
+            qry = qry.filter(Bmark.is_private == False)    # noqa
```
mitechie, line 405: can you comment on the required # noqa ?

bookie/tests/test_api/test_popular_api.py
```diff
@@ -45,7 +45,7 @@ def _check_cors_headers(self, res):
         self.assertEqual(
             res.headers['access-control-allow-headers'], 'X-Requested-With')
 
-    def _add_bookmark(self, user=None):
+    def _add_bookmark(self, user=None, is_private=False):
```
mitechie: hmm, so the other helpers default to is_private to true correct? I'd think we want to keep these consistent to aid in test writing across code.

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -264,3 +264,70 @@ def test_delete_all_bookmarks(self):
                 'There should be no tags left: ' + str(len(tags))
             )
             DBSession.flush()
+
+    def test_find_bookmarks_same_user(self):
+        """Testing that if a user requests for their own bookmarks, they
```
mitechie: This should be one line into and then more details. Check out pep257 in python land.

```
""""A user requesting their bookmarks includes private ones"""
```

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -264,3 +264,70 @@ def test_delete_all_bookmarks(self):
...
+        res = BmarkMgr.find(username=user.username, requested_by=user.username)
+        self.assertEqual(
+            bookmark_count,
+            len(res),
+            'There should be ' + str(bookmark_count) + ' bookmarks present: ' +
+            str(len(res))
+        )
+
+    def test_find_bookmarks_diff_user(self):
+        """Testing that if a user requests for another's bookmarks, they
```
mitechie: A user requesting another user's bookmarks get public only"

bookie/tests/test_models/test_bmarkmgr.py
```diff
@@ -264,3 +264,70 @@ def test_delete_all_bookmarks(self):
                 'There should be no tags left: ' + str(len(tags))
             )
             DBSession.flush()
+
+    def test_find_bookmarks_same_user(self):
+        """Testing that if a user requests for their own bookmarks, they
+        should get all the bookmarks back"""
+        bookmark_count = 10
```
mitechie: this should have the same setup as the other test. Some public, some private. In this way we verify there's no bug in getting a user their private OR public bookmarks.

mitechie:
Generally looks good with a few notes to look into.

bookie/models/__init__.py
sambuddhabasu, line 405: If # noqa is not used, I get the error:
comparison to False should be 'if cond is False:' or 'if not cond:'

bookie/tests/test_api/test_popular_api.py
sambuddhabasu: These tests were failing before but now run fine without any changes, so, I have removed the extra is_private parameter and the bookmarks are being saved as private by default.

bookie/models/__init__.py
mitechie, line 405: Please add that as a comment for other devs in the code please.

mitechie:
Please add the comment and then :+1:

preddy5:
:+1:

sambuddhabasu:
:shipit:

PR #476, opened 2014-06-16
Update api and corresponding models

sambuddhabasu:


bookie/models/__init__.py
```diff
@@ -184,8 +185,16 @@ def complete(prefix, current=None, limit=5, username=None):
...
 
             # if we have a username limit to only bookmarks of that user
-            if username is not None:
+            if username:
                 qry = qry.filter(Tag.bmark.any(username=username))
+                if username != requested_by:
+                    bmark = aliased(Bmark)
+                    qry = qry.join((bmark, Tag.bmark)).\
+                        filter(bmark.is_private == False)   # noqa
+            else:
```
craigmaloney, line 198: What happens if username == requested by? Looks like this falls through.

sambuddhabasu, line 198: if username == requested_by, we don't need to filter on anything, as in that case, we would want both the private and the public bookmarks. We already get both public and private bookmarks by the above line of code, qry = qry.filter(Tag.bmark.any(username=username))

bookie/tests/test_models/test_tagmgr.py
```diff
@@ -24,27 +26,186 @@ def test_total_ct(self):
...
+
+    def test_basic_complete_same_user_accounts_for_privacy(self):
+        """Tags should provide completion options."""
+        user = User()
+        user.username = gen_random_word(10)
+        DBSession.add(user)
+
+        # Generate demo tag into the system
+        tags = [make_tag() for i in range(5)]
+        [DBSession.add(t) for t in tags]
```
mitechie, line 74: I thought we were going to take these out and just pass the ' '.join(tags) into the BmarkMgr.store() ?

mitechie:
One comment on making the tests simpler and then this looks good to go. Thanks for the hard work. :+1:

bookie/tests/test_models/test_tagmgr.py
sambuddhabasu, line 74: The BmarkMgr.store() has been replaced with make_bookmark() now, so we would need to add the tags.

preddy5:
:+1:

mitechie:
:+1:

sambuddhabasu:
:shipit:

PR #479, opened 2014-06-18
twitter oauth

preddy5:


preddy5:
@mitechie could you review this PR, Thank You.

mitechie:
@pradyu1993 please provide more details and information in your pull requests. You have a title that's very generic, no details on what's changed and why in the code, no instructions for testing out and QA'ing the change. Those things are vital to code reviewers to doing a nice and helpful code review.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,55 @@
...
+
+<html>
+<title>Bookie - Connecting...</title>
+
+<head>
+    <meta name="viewport" content="initial-scale=1, maximum-scale=1">
+    
+    <style>
+        body {
+            font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
```
mitechie: the CSS shouldn't be in here but in the sass. This isn't using any of the fonts used in bookie currently and we need to match that up.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,55 @@
...
+            font-size: 16px;
+            font-weight: bold;
+            color: #A0A0A0;
+            font-style: italic;
+        } 
+    </style>
+    <script type="text/javascript" charset="utf-8">
```
mitechie: all JS should be in a pulled out module.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,55 @@
...
+
+        if (!window.console) {
```
mitechie: you shouldn't have to worry about this. Was there a problem this addressed?

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,55 @@
...
+            window.console = {};
+            window.console.log = function() {};
+        }
+        %if url:
+            var next = "${url}";
+            console.log(["Forwarding to next", next]);
+            setTimeout(function() {
```
mitechie, line 18: there must be a better way than a timeout to go through this. Timeouts are hard to test and very fragile.

bookie/views/social.py
```diff
@@ -0,0 +1,61 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+@view_config(route_name="twitter_connect",
+             renderer="/social/social_connect.mako")
+def twitter_connect(request):
+    """function to get users twitter oauth information"""
+
+    twitter_consumer_key = 'i1nxL7ViGSkedntSmMmcYiKA3'
```
mitechie: these need to be pulled from somewhere? Are these for the app or the user? If they're for the app it should be part of the ini. If the user, we should be getting a user object and doing this.

bookie/views/social.py
```diff
@@ -0,0 +1,61 @@
...
+                error('Twitter connection denied tweepy IOError')
+            return dict(result="""Twitter has returned an
+             error. Try connecting again.""")
+
+        connections = TwitterConnection.query.filter(
+            TwitterConnection.uid == unicode(twitter_user.id)).all()
+        if connections and connections[0].username == request.user.username:
+            return dict(result="""Another user (%s) has already connected
+             with those Twitter credentials.""" % (request.user.username))
+        tconnection = TwitterConnection()
```
mitechie: we should be able to to do something where we pass the args in to the new connection and not need to list it all out like this.

mitechie:
This seems a good start but I'd appreciate some help on how this is supposed to work, some way to test some parts of this, and some comments/etc in the code itself.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,48 @@
...
+            rel='stylesheet' type='text/css'/>
+        <link rel="stylesheet" type="text/css" href="/static/css/responsive.css"/>    
+
+        <script type="text/javascript" charset="utf-8">
+        %if url:
+        <!-- If Credentials are not present in the URL this timeout redirects to twitter oauth url to recieve users oauth details -->
+            var next = "${url}";
+            console.log(["Forwarding to next", next]);
+            setTimeout(function() {
+                console.log(["Forwarding to next", next]);
```
mitechie, line 19: Let's not have 2 places with the same console.log as you can't tell which one was running.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,48 @@
...
+                window.location.href = next;
+            }, 1000); 
+        %else:
+        <!-- Once Credentials are received this timeout redirects to settings page closing the windows-->
+            setTimeout(function() {
+                console.log("${result}");
+                window.close();
+                window.location.href = "/";
```
mitechie: this needs to be the root url passed in to the route vs just hard coded at / to work when bookie is hosted in a subdirectory. Check out the tools in pyramid to build a route.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,48 @@
...
+            }, 2000);
```
mitechie, line 28: why is this one 2s and the other one 1s?

bookie/views/social.py
```diff
@@ -0,0 +1,71 @@
...
+        request.registry.settings.get('twitter_consumer_key', False)
+    twitter_consumer_secret = \
+        request.registry.settings.get('twitter_consumer_secret', False)
+
+    oauth_token = request.params.get('oauth_token', None)
+    oauth_verifier = request.params.get('oauth_verifier', None)
+    denied = request.params.get('denied', None)
+    if denied:
+        LOG.error('Twitter connection denied')
+        return {'result': 'Denied! Try connecting again.'}
```
mitechie: Let's make this a bit nicer.

```
Access to Twitter denied. Please try connecting again
```

And let's make sure we give them a link or something to try again.

bookie/views/social.py
```diff
@@ -0,0 +1,71 @@
...
+            twitter_user = api.me()
+        except (tweepy.TweepError, IOError):
+            LOG.\
+                error('Twitter connection denied tweepy IOError')
+            return dict(result="""Twitter has returned an\
+             error. Try connecting again.""")
+
+        connections = TwitterConnection.query.filter(
+            TwitterConnection.uid == unicode(twitter_user.id)).all()
+        if connections and connections[0].username == request.user.username:
```
mitechie, line 52: why do we need to check both the connections and the user? If the connection was found, twitter uid was found, then there's a duplicate?

sample.ini
```diff
@@ -21,6 +21,10 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_index
 
+# twitter application details
+twitter_consumer_key = i1nxL7ViGSkedntSmMmcYiKA3
```
mitechie: Are these your legit credentials? We should add place holders here.

test.ini
```diff
@@ -21,6 +21,10 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_test
 
+# twitter application details
```
mitechie, line 24: same here, the tests should not talk to twitter directly but be mocked out.

test_mysql.ini
```diff
@@ -20,6 +20,10 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_mysql
 
+# twitter application details
+twitter_consumer_key = i1nxL7ViGSkedntSmMmcYiKA3
```
mitechie: same here

test_pgsql.ini
```diff
@@ -20,6 +20,10 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_pgsql
 
+# twitter application details
+twitter_consumer_key = i1nxL7ViGSkedntSmMmcYiKA3
```
mitechie: same here

mitechie:
Thanks for the update. This is looking better. I've added a few comments. We also need to think about testing this. We need to collect the responses from twitter and use mocking to help us write tests for the code that's written here by pretending to make the api calls and get the correct responses for both errors and success.

bookie/views/social.py
preddy5, line 52: first time when table is empty connection[0] will raise error

preddy5:
@mitechie could you please look at the tests I have tried my best but am unable to get them working

bookie/lib/social_utils.py
```diff
@@ -0,0 +1,19 @@
+
```
mitechie, line 1: So looking at this I'll look for a test_social_utils.py

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,50 @@
...
+<head>
+    <meta name="viewport" content="initial-scale=1, maximum-scale=1">
+    
+        <link
+            href='https://fonts.googleapis.com/css?family=Cabin|Cabin+Sketch:bold&v2'
+            rel='stylesheet' type='text/css'/>
+        <link rel="stylesheet" type="text/css" href="/static/css/responsive.css"/>    
+
+        <script type="text/javascript" charset="utf-8">
+        %if url:
```
mitechie, line 15: I'm also looking for a test where this view is called with a url and without to make sure things are setup correctly.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,50 @@
...
+                window.location.href = "${request.route_url('user_account', username=request.user.username)}";
+            }, 2000);
+        %endif
+    </script>
+</head>
+
+<body>
+
+    <div style="text-align: center;">
+        %if result:
```
mitechie, line 36: as well as one with and without result to make sure the right information is output to the user.

bookie/templates/social/social_connect.mako
```diff
@@ -0,0 +1,50 @@
...
+            <h2>${result}</h2>
+            %if retry_link:
+            <a href="${request.route_url('twitter_connect')}">Click Here</a>
+            %endif
+        %else:
+            <img src="/static/images/play_store_512.png" style="padding: 0 5px;" alt="Bookie Extensions"/>    
+            <div>Hold on just a sec...</div>
+            <div>If I was a language I'd be python!!!</div>
```
mitechie: This doesn't provide any useful info to the user as to what is going on?

bookie/tests/test_lib/save.p
```diff
@@ -0,0 +1,423 @@
+ccopy_reg
```
mitechie: what is this?  This file should be removed I think.

bookie/tests/test_lib/test_social_utils.py
```diff
@@ -0,0 +1,41 @@
...
+from bookie.tests import TestViewBase
+from bookie.lib.social_utils import (
+    create_twitter_userapi,
+    create_twitter_OAuthHandler)
+
+
+class TestSocialOAuth(TestViewBase):
+
+    def test_search_content(self):
+        """Test if correct urls are rendered if OAuthHandler
```
mitechie: this should fit on one line right?

bookie/tests/test_lib/test_social_utils.py
```diff
@@ -0,0 +1,41 @@
...
+import transaction
+
+from bookie.tests import factory
+from bookie.tests import TestViewBase
+from bookie.lib.social_utils import (
+    create_twitter_userapi,
+    create_twitter_OAuthHandler)
+
+
+class TestSocialOAuth(TestViewBase):
```
mitechie: nothing in test_list should have to extend TestViewBase. The lib directory isn't for views.

bookie/tests/test_lib/test_social_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_search_content(self):
+        """Test if correct urls are rendered if OAuthHandler
+        is edited"""
+        create_twitter_OAuthHandler = mock.Mock(
```
mitechie: So this should be a test_create_twitter_userapi. It should verify that given a consumer key/secret, and oauth token/verifier, (mocked or made up or whatever) it returns an api instance and a access token.

bookie/tests/test_lib/test_social_utils.py
```diff
@@ -0,0 +1,41 @@
...
+            return_value="https://twitter_url/")
+        self._login_admin()
+        res = self.app.get('/oauth/twitter_connect')
+        self.assertTrue(
+            "https://twitter_url/" in res.body,
+            msg="OAuth should render url returned by\
+             OAuthHandler when no params")
+
+    def test_credentials_duplicate(self):
```
mitechie: and I'd expect this to be a test of the other function in the lib/social_utils.py create_twitter_oauthhandler.

bookie/tests/test_lib/test_social_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        message = """Another user (Admin) has already connected\
+             with those Twitter credentials."""
+        factory.make_twitter_connection()
+        transaction.commit()
+        twitter_user = pickle.load(
+            open("bookie/tests/test_lib/save.p", "rb"))
+        create_twitter_userapi = mock.Mock(return_value=twitter_user)
+        self._login_admin()
+        res = self.app.get('/oauth/twitter_connect',
+                           status=200)
```
mitechie: These are view based so I'd expect this to be in a test_webviews/test_social.py

bookie/views/social.py
```diff
@@ -0,0 +1,72 @@
...
+            # If any other user has provided the same credentials message is
+            # shown
+            return dict(result="""Another user (%s) has already connected\
+             with those Twitter credentials.""" % (request.user.username))
+        tconnection = TwitterConnection(
```
mitechie: The views should not talk to the database or database models directly. They should go through a manager like the BmarkMgr.store method. In this way views are kept apart from dealing with database code and the db code can be mocked our for testing very easily. Replace the SocialMgr.save (or whatever you want to call it) and you can test this view easier.

sample.ini
```diff
@@ -21,6 +21,10 @@ email.host=sendmail
 fulltext.engine=whoosh
 fulltext.index=bookie_index
 
+# twitter application details
+twitter_consumer_key = Guesswhat
+twitter_consumer_secret = BookieRocks
```
mitechie, line 26: :)

mitechie:
Thanks for the hard work prady. This is getting closer. I've got a few comments on some missing tests, moving a few tests, and some more cleanup of database calls.

preddy5:
@mitechie could you please look into the tests of social_utils

mitechie:
Prady, I've poked at the branch and the tests. I agree that the lib/social stuff can go without the tests. It's all tweepy stuff. 

The other tests still need some work. I added two initial tests and did some refactoring of the view/social code. It's over in https://github.com/bookieio/Bookie/pull/490/files

Please give it a look over and see if you that makes sense and helps provide some direction towards adding some more tests to the view code.

PR #481, opened 2014-06-25
Update fulltext search to take care of private bookmarks

sambuddhabasu:


bookie/bcelery/tasks.py
```diff
@@ -255,6 +255,8 @@ def fulltext_index_bookmark(bid, content):
                 extended=b.extended if b.extended else u"",
                 tags=b.tag_str if b.tag_str else u"",
                 readable=found_content,
+                username=b.username,
```
mitechie, line 258: this change should have an updated test to detect and make sure the new fields are there.

bookie/models/fulltext.py
```diff
@@ -10,11 +10,21 @@
 from sqlalchemy.orm import joinedload
 
 from whoosh import qparser
-from whoosh.fields import SchemaClass, TEXT, KEYWORD, ID
+from whoosh.fields import (
+    BOOLEAN,
```
mitechie, line 14: thanks for cleaning that up

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +677,74 @@ def test_search_api(self):
...
+            '/api/v1/admin/bmarks/search/google',
+            status=200
+        )
+
+        # Make sure we can decode the body.
+        bmark_list = json.loads(res.body)
+        results = bmark_list['search_results']
+        self.assertEqual(
+            len(results),
+            1,
```
mitechie: So this test isn't great because you're verifying that you only got one result back, but there's only one bmark in the db. A great test would be to add a second bmark by a second user and making sure only one came out.

sambuddhabasu: A different test called test_search_api_diff_user has been added to make sure that a user can get only their bookmarks on search.

mitechie: That's awesome that there's a new test, but this test could completely pass while being broken. Ideally the test should be able to fail in one obvious way and if anything goes wrong.

mitechie:
Really close, I think the query needs a small update and then tests to cover all the cases. 
- Load up two users, user 1 loads two public and one private bookmark. User two loads 1 private and 1 public bookmark. User 1 does a query and gets a total of 4 bookmarks. His public/private and user 2's public one.
- Same setup, but as as user 1 requesting only their own bookmarks. In this case they should only get 3 bookmrks.
- Same setup but as user a third user requesting user 1's bookmarks. He should only get two bookmarks.

bookie/models/fulltext.py
```diff
@@ -90,6 +104,17 @@ def search(self, phrase, content=False, username=None, ct=10, page=0):
                                               group=qparser.OrGroup)
             qry = parser.parse(phrase)
 
+            if username:
+                if requested_by and username == requested_by:
+                    qry = And([qry, Or([Term('is_private', 'f'),
```
mitechie: This looks good, can we explode it up a bit to make it more clear about what's getting built. I know it'll take up more lines but the hope is that it'd be more clear when reading the code. 

For instance this reads a bit more clear to me:

```
qry = And([
    qry,
    Or([
        Term('is_private', 'f'),
        And([
           Term('username', username),
           Term('is_private', 't')
        ])
    ])
])
```

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,158 @@ def test_search_api(self):
...
+        url_four = u'http://google4.com'
+        # Make one public bookmark for admin.
+        self._get_good_request(url=url_one)
+        # Make one private bookmark for admin.
+        self._get_good_request(url=url_two, is_private=True)
+
+        # Make a test user.
+        test_user_username = u'test_user'
+        test_user = factory.make_user(username=test_user_username)
+        test_user_apikey = test_user.api_key = u'random_key'
```
mitechie: why do we need both test_user_apikey and test_user.api_key? Can we not just reference test_user.api_key?

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,158 @@ def test_search_api(self):
...
+            format(len(results)))
+        self.assertTrue(
+            results[0][u'username'] == u'admin' and not
+            results[0][u'is_private'] and results[0][u'url'] == url_one,
+            "We should have a public bookmark from admin")
+        self.assertTrue(
+            results[1][u'username'] == u'admin' and
+            results[1][u'is_private'] and results[1][u'url'] == url_two,
+            "We should have a private bookmark from admin")
+        self.assertTrue(
```
mitechie: Awesome, this is exactly what the test should be checking. <3

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,158 @@ def test_search_api(self):
...
+            results[0][u'username'] == u'admin' and not
+            results[0][u'is_private'] and results[0][u'url'] == url_one,
+            "We should have a public bookmark from admin")
+        self._check_cors_headers(res)
+
+    def test_search_api_anon_user(self):
+        """Test that an anonymous user gets only public bookmarks"""
+        url_one = u'http://google1.com'
+        url_two = u'http://google2.com'
+        url_three = u'http://google3.com'
```
mitechie: this setup is the same across tests. Let's move this into a common helper like the _get_good_request that sets up both users and their bookmarks. Then the tests are just about the checks. I'd rename the url_one and url_two to be more like:

```
admin_public_url = ...
admin_private_url = ...
user_public_url = ...
user_private_url = ...
```

And then the tests will read nice and clear that we assert we have admin's public and private url and user's public url, etc.

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,158 @@ def test_search_api(self):
...
+        self._get_good_request(url=url_three, username=test_user_username,
+                               api_key=test_user_apikey)
+
+        # Make one private bookmark for test user.
+        self._get_good_request(url=url_four, username=test_user_username,
+                               api_key=test_user_apikey, is_private=True)
+
+        # Search for bookmark as admin.
+        res = self.testapp.get(
+            '/api/v1/bmarks/search/google',
```
mitechie, line 805: this is not a search as admin as the comment states.

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,158 @@ def test_search_api(self):
...
+        bmark_list = json.loads(res.body)
+        results = bmark_list['search_results']
+        self.assertEqual(
+            len(results),
+            2,
+            "We should have two results coming back: {0}".
+            format(len(results)))
+        self.assertTrue(
+            results[0][u'username'] == u'admin' and not
+            results[0][u'is_private'] and results[0][u'url'] == url_one,
```
mitechie: this is where it would read a lot nicer. We'd be asserting that we have admin_url_public and user_url_public

bookie/tests/test_utils/test_fulltext.py
```diff
@@ -36,7 +36,7 @@ def tearDown(self):
         testing.tearDown()
         empty_db()
 
-    def _get_good_request(self, new_tags=None):
+    def _get_good_request(self, new_tags=None, is_private=False):
```
mitechie, line 39: You added the is_private to false here but I don't see that used in any of the new tests/code? Is this a required change or just done for some other reason?

mitechie:
This looks really really good and addresses the changes. I've got some gardening for you before landing but this looks great and Craig can take one last peek on the refactor of the tests once that's updated.

Thanks for pushing through this. I'm really happy with where the code is ending up.

sambuddhabasu:
Thank you so much @mitechie, I will be updating the code so that it reads clear and nice.

bookie/tests/test_utils/test_fulltext.py
sambuddhabasu, line 39: Yes, this is a required change, as in the tests, no api_key is being passed so, we would want the bookmarks to be public by default so that we can fetch them even without providing an api_key.

bookie/tests/test_api/test_base_api.py
```diff
@@ -677,6 +684,229 @@ def test_search_api(self):
...
+        }
+        user_private_bmark = {
+            'url': u'http://google4.com',
+            'username': test_user.username,
+            'api_key': test_user.api_key,
+            'is_private': True,
+        }
+
+        bmark_test = {}
+        bmark_test['admin_public_bmark'] = admin_public_bmark
```
craigmaloney: You could trim these down even further:

bmark_test['admin_public_bmark'] = {
             'url': u'http://google1.com',
             'username': u'admin',
             'api_key': API_KEY,
             'is_private': False}

bookie/tests/test_api/test_base_api.py
```diff
@@ -132,6 +139,50 @@ def _setup_user_bookmark_count(self):
...
+        }
+
+        # Make the bookmarks.
+        for bmark in bmark_test.keys():
+            self._get_good_request(url=bmark_test[bmark]['url'],
+                                   username=bmark_test[bmark]['username'],
+                                   api_key=bmark_test[bmark]['api_key'],
+                                   is_private=bmark_test[bmark]['is_private'])
+
+        return [test_user, bmark_test]
```
craigmaloney: Maybe use a tuple here?

bookie/tests/test_api/test_base_api.py
craigmaloney: I think the make_test_bookmarks now adds more than one bookmark so this should alleviate the concern.

craigmaloney:
Looks good.

Not sure if this will work or what it will do, but here goes:

:shipit:

preddy5:
:+1:

sambuddhabasu:
:shipit:

PR #485, opened 2014-07-07
Update fulltext search to support private bookmark

sambuddhabasu:
Added username and is_private to fulltext index celery task

Query in a smarter way

Users not logged in should get only public bookmarks on search

Added api tests for fulltext search

Added tests

Added fulltext schema test

Added fulltext search test

Added more api tests

Update fulltext search api tests

Tests look nicer now

Tests are modified

Created new function for tests

Return data type changed


sambuddhabasu:
This pull request is in place of #481 .

sambuddhabasu:
:shipit:

PR #486, opened 2014-07-07
Update user export

sambuddhabasu:


mitechie:
Looks good but no tests so will wait to formally review it and qa it.

bookie/templates/utils/export.mako
```diff
@@ -10,7 +10,7 @@
     % for bmark in bmark_list:
         <DT>
             <A HREF="${bmark.hashed.url}" LAST_VISIT="" ADD_DATE="${time.mktime(bmark.stored.timetuple())}"
-               TAGS="${','.join([tag for tag in bmark.tags])}">
+               TAGS="${','.join([tag for tag in bmark.tags])}" IS_PRIVATE="${bmark.is_private}">
```
mitechie: If you look at the test files we use for import testing, the key is just PRIVATE and not IS_PRIVATE.

bookie/tests/test_utils/test_export.py
```diff
@@ -84,3 +85,33 @@ def test_export_wo_tags(self):
...
+        self._get_good_request()
+
+        res = self.app.get('/admin/export', status=200)
+
+        self.assertTrue(
+            "google.com" in res.body,
+            msg='Google is in the exported body: ' + res.body)
+
+        self.assertTrue(
+            'IS_PRIVATE="False"' in res.body,
```
mitechie: so then this assertion needs to be updated

bookie/tests/test_utils/test_export.py
```diff
@@ -84,3 +85,33 @@ def test_export_wo_tags(self):
...
+        self._get_good_request(is_private=True)
+
+        res = self.app.get('/admin/export', status=200)
+
+        self.assertTrue(
+            "google.com" in res.body,
+            msg='Google is in the exported body: ' + res.body)
+
+        self.assertTrue(
+            'IS_PRIVATE="True"' in res.body,
```
mitechie: as does this one.

bookie/tests/test_utils/test_export.py
mitechie: and in the false case, we should just not apply the private attribute to the item. It should only show if it is.

bookie/tests/test_utils/test_export.py
mitechie: also note that the attribute is just a 1

```
PRIVATE="1"
```

mitechie:
The attribute it a bit off of what I think it needs to be so that it's the proper format. Please also verify you can import your output into the system as well and get a private bookmark in the system.

mitechie:
:+1: thanks for the update.

preddy5:
:+1:

sambuddhabasu:
:shipit:

PR #487, opened 2014-07-09
Update user import

sambuddhabasu:


mitechie:
:+1: Thanks for the update.

bookie/lib/importer.py
```diff
@@ -181,9 +182,9 @@ def process(self):
 
             link = tag.a
 
-            # Skip any bookmarks with an attribute of PRIVATE.
-            if link.has_key('PRIVATE'):  # noqa
-                continue
+            is_private = False
+            if link.has_key('private'):  # noqa
```
craigmaloney, line 186: Wondering if we should do a search in the keys to cover "private", "is_private" or any other version of this.

That said, maybe that's more trouble than its worth.

craigmaloney:
:+1:

bookie/lib/importer.py
sambuddhabasu, line 186: For DelImporter, the private bookmarks are marked with a 'PRIVATE="1"', so we are searching here with key 'private'.

sambuddhabasu:
:shipit:

PR #493, opened 2014-07-26
Feature runit for webservice

ebeckm:
An example runit file for keeping a web server running.  Helpful, for example, when your hosting provider periodically reboots your machine.


ebeckm:
Tested on Ubuntu 14.04

runit/paster-service/log/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec chpst -u root svlogd -tt /home/bookie/Bookie/runit/paster-service/logs
```
mitechie, line 2: this seems odd that it's got a full path defined. Isn't this very not portable?

runit/paster-service/log/supervise/pid
```diff
@@ -0,0 +1 @@
+12618
```
mitechie, line 1: should the log/pid be part of this commit? That pid will change each time it's run right?

mitechie, line 1: the lock file above seems to fit into the same category.

runit/paster-service/run
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+exec 2>&1
+set -e
+
+PASTER=/home/bookie/Bookie/bin/paster
```
mitechie, line 5: this seems the meat of what is required. The paths I think should be generated via either the makefile or something that knows the path vs requiring all other users to use the same path.

mitechie:
Thanks for this. Sorry, I've been traveling so my reply is a bit late. I think it's great to have an runit implementation there. I think it might be best to look at this as something a Makefile target could generate on the fly. In this way, you don't have to have it if you're not using runit, but when you do need it, you can generate it as required with the proper paths/etc all setup for you. 

Let me know if you need a hand with any Make magic or if I'm mis-understanding how this is meant to work.

PR #494, opened 2014-07-31
Private indication for recent bookmarks

sambuddhabasu:


preddy5:
:+1:

mitechie:
We should have a test on the view that it's added properly and not added when it shouldn't. Once that is set this is ok.

craigmaloney:
:+1:

PR #495, opened 2014-08-02
twitter celery job

preddy5:


bookie/bcelery/tasks.py
```diff
@@ -50,10 +54,12 @@ def daily_stats():
 
     Currently we're monitoring:
     - Total number of bookmarks for each user in the system
-
+    - Delete's inactive accounts if any
+    - Refresh's Twitter fetch from user's accounts
     """
     count_total_each_user.delay()
     delete_non_activated_account.delay()
+    refresh_twitter_fetch.delay()
```
mitechie: I'm not sure about this placement. The function is named daily stats, but this isn't stats related.

bookie/bcelery/tasks.py
```diff
@@ -285,6 +291,21 @@ def reindex_fulltext_allbookmarks(sync=False):
 
 
 @celery.task(ignore_result=True)
+def refresh_twitter_fetch(username=None):
+    oauth_token = INI.get('twitter_consumer_key')
+    oauth_verifier = INI.get('twitter_consumer_secret')
+    for connection in SocialMgr.get_twitter_connections(username):
```
mitechie, line 304: so we should have a task that fetches the users with social connections and calls a separate function for each user in the celery queue. In this way each user processing is an item in the queue and workers could be processing more than one user's twitter feed at once.

bookie/models/social.py
```diff
@@ -36,6 +46,13 @@ def store_twitter_connection(username, credentials):
         DBSession.add(tconnection)
         return tconnection
 
+    @staticmethod
+    def update_lasttweet_data(connection, tweet_id):
```
mitechie: update_last_tweet_data

bookie/tests/test_api/test_social_api.py
```diff
@@ -32,3 +33,21 @@ def testSocialAuthentication(self):
         self.assertEqual(
             res.status, "403 Forbidden",
             "status should be 403")
+
+    @patch('bookie.bcelery.tasks.refresh_twitter_fetch')
+    def testTwitterRefresh(self, mock_refresh):
+
```
mitechie: no blank line here.

dbversions/versions/dbc7a0f1182_adding_last_twitter_id_column.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = 'dbc7a0f1182'
+down_revision = '352bb5f4fff9'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('TwitterConnection', sa.Column('last_tweet_id', sa.Unicode(length=255), nullable=True))
```
mitechie: What do you think of calling this 'last_tweet_seen' instead?

bookie/bcelery/tasks.py
preddy5: What do you think I should do? Shall I add a separate function just for scheduling twitter bot?

mitechie: Yes please. I think this makes sense to be able to enable/schedule the twitter processing apart from stats.

bookie/bcelery/tasks.py
```diff
@@ -359,3 +380,34 @@ def fetch_bmark_content(bid):
...
+    tweets = twitter_user.user_timeline(
+        id=connection.twitter_username,
+        include_entities=True,
+        since_id=connection.last_tweet_seen)
+    if tweets:
+        SocialMgr.update_last_tweet_data(connection, tweets[0].id)
+        for tweet in tweets:
+            for url in tweet.entities['urls']:
+                expanded_url, title = get_url_title(url['expanded_url'])
+                BmarkMgr.store(
```
mitechie: we should check if the url/tweet is already bookmarked I think.

bookie/bcelery/tasks.py
```diff
@@ -359,3 +380,34 @@ def fetch_bmark_content(bid):
...
+
+
+@celery.task(ignore_result=True)
+def fetch_tweets(twitter_user, connection):
+    tweets = twitter_user.user_timeline(
+        id=connection.twitter_username,
+        include_entities=True,
+        since_id=connection.last_tweet_seen)
+    if tweets:
+        SocialMgr.update_last_tweet_data(connection, tweets[0].id)
```
mitechie: maybe update the last tweet status after processing in case something goes wrong in that for loop.

bookie/bcelery/tasks.py
```diff
@@ -285,6 +297,15 @@ def reindex_fulltext_allbookmarks(sync=False):
 
 
 @celery.task(ignore_result=True)
+def refresh_twitter_fetch(username=None):
+    """
+    Run twitter fetch for required username's
+    """
+    for connection in SocialMgr.get_twitter_connections(username):
+        create_twitter_api(connection)
```
mitechie, line 305: how about naming this 'process_twitter_connections' or something?

bookie/routes.py
```diff
@@ -205,6 +205,14 @@ def build_routes(config):
         "api_admin_imports_reset",
         "/api/v1/a/imports/reset/{id}",
         request_method="POST")
+    config.add_route(
```
mitechie, line 208: this is great. Thanks for this. We need to update the docs/api.rst with these calls in the admin section as well.

bookie/models/social.py
```diff
@@ -36,6 +46,13 @@ def store_twitter_connection(username, credentials):
         DBSession.add(tconnection)
         return tconnection
 
+    @staticmethod
+    def update_last_tweet_data(connection, tweet_id):
+        connection.last_tweet_seen = tweet_id
+        connection.refresh_date = datetime.now()
+        DBSession.add(connection)
```
mitechie: you don't need to .add an existing object that came from the db. It should just update at the end of the request.

bookie/tests/test_models/test_socialmgr.py
```diff
@@ -1,11 +1,23 @@
...
 
 from bookie.models.social import SocialMgr
-from bookie.tests import TestDBBase
+from bookie.tests import (
+    factory,
+    empty_db)
 
 
-class TestSocialMgr(TestDBBase):
+class TestSocialMgr(TestCase):
```
mitechie, line 13: why did you remove the TestDBCase parent that makes sure to reset the db automatically?

mitechie:
Thanks for the updates. This is looking really good. I really like the tests. I think we could use a couple around the celery side as well. There should be come celery tests that you can look at to see about testing the new celery functions.

bookie/tests/test_models/test_socialmgr.py
preddy5, line 13: https://github.com/bookieio/Bookie/blob/develop/bookie/tests/__init__.py#L80 will raise a error because twitterconnection has a foreign key so empty_db() should be run while tearDown not abort

mitechie:
Thanks for the updates. :+1:

craigmaloney:
:+1:

sambuddhabasu:
:+1:

preddy5:
:shipit:

PR #501, opened 2014-08-30
Social frontend

preddy5:


anarcat:
before implementing logins only for one social network, why not consider using something like https://indieauth.com/ or persona?

PR #502, opened 2014-09-04
Document sqlite -> mysql/postgresql migration and nltk_data environment variable in wsgi

jcharaoui:
I found it useful to pass the NLTK_DATA environment variable when hosting Bookie in wsgi, otherwise I would get a MissingCorpusException. Other than the NLTK_DATA envvar, the library seems to look for the nltk data only in a handful of hardcoded paths.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -122,6 +122,44 @@ the database for you. Once you're ready run:
 
     $ make db_up
 
+
+Migrate from SQLite to MySQL or Postgresql
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+First, follow the steps above to set up an empty MySQL/Postgresql database.
+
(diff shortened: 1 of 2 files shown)


docs/install.rst
```diff
@@ -122,6 +122,44 @@ the database for you. Once you're ready run:
...
+    $ apt-get install ruby-sqlite3 ruby-pg
+
+    $ gem install rack -v 1.4.5
+    $ gem install taps
+
+Next, let's publish our existing sqlite database:
+
+::
+
+    $ taps server sqlite:///home/user/bookie/bookie/bookie.db tmpuser tmppass &
```
mitechie, line 153: cool had never heard of taps before and this is kind of a cool use.

mitechie:
Thanks for these doc updates. Really appreciated.

PR #512, opened 2014-09-13
Add support for finding/fixing missing fulltext

mitechie:
- Add fulltext numbers to the dashboard to help identity problems.
- Add a celery job to search for bookmarks missing from fulltext and add them
- Celery job grabs 500 bookmarks every 10minutes.


mitechie:
Tests pass

The update to JS will require an increment to the JS url counter in the bookie.ini.

PR #513, opened 2014-09-18
Update export to take into account privacy

mitechie:


bookie/tests/test_utils/test_export.py
```diff
@@ -52,6 +52,7 @@ def test_export(self):
         """Test that we can upload/import our test file"""
         self._get_good_request()
 
+        self._login_admin()
```
mitechie, line 55: The requests now need a logged in user to export.

Issue #520, opened 2014-10-28
bookmark export is no sorting the bookmarks by time


alexlehm:
I would prefer if the export bookmarks file would be sorted by time. I initially thought the export was missing some bookmarks, but it turned out that they were somewhere in the middle.

mitechie:
Thanks, it should be a simple sort on export. I'll look at getting that updated.

PR #524, opened 2014-11-02
fixed issue #520 - sorting bookmarks in descending order while exporting

AnishShah:
Hey @mitechie , I wrote the test case which adds bookmark on random date. Without my patch, the json response is not sorted. Please check.


mitechie:
Thanks for this, can you run make lint and make sure that it passes? A few lines look a bit long.

AnishShah:
@mitechie There were some small PEP8 errors. I have fixed it. `make lint` passes now.

mitechie:
Lint and tests pass. Thanks for the update!

Issue #528, opened 2014-11-16
tags are matched with case, while the search doesn't


alexlehm:
I'm not sure if this is intentional or not, I had some problems finding one of my old bookmarks since I had entered the tag with a capital letter. The tag search returns the lower case tag, but the search then uses the capitalized one.

works: https://bookie.io/alexlehm/recent/discourse?count=50&page=0
no results: https://bookie.io/alexlehm/recent/Discourse?count=50&page=0

The keyword search is case insensitive however.

yashmehrotra:
Hi , I am new to Bookie.I love your app and would like to contribute to it.
I have fixed this bug and tested it. It works.
Please review my pull request.
https://github.com/bookieio/Bookie/pull/530

yashmehrotra:
Here is the screenshot of the fixed issue-
![screenshot from 2014-11-20 17 04 14](https://cloud.githubusercontent.com/assets/5863972/5123950/66df4aaa-70d7-11e4-8d93-99fe14b366b7.png)

![screenshot from 2014-11-20 17 04 29](https://cloud.githubusercontent.com/assets/5863972/5123955/6da5e5e2-70d7-11e4-8382-bb402bf7f4d0.png)

PR #530, opened 2014-11-20
Fix of Issue #528

yashmehrotra:
I tested the case matching and it works

### DIFF
--- a/bookie/tests/test_api/test_base_api.py
+++ b/bookie/tests/test_api/test_base_api.py
@@ -1294,3 +1294,28 @@ class BookieAPITest(unittest.TestCase):
         self.assertTrue(
             'with_bookmarks' in data,
             "Should have count of users with bookmarks: " + str(data))
+            
+    def test_tag_search_casing(self):
+        """Test that search works same with all types of casing"""
+
+        # Adding the tags 'python' and 'search'
+        self._get_good_request()
+
+        # Testing the tag 'python' and 'search' with different cases
+        word_list = ['python', 'Python', 'pYthon', 'PYTHoN', 'pythON',
+                     'PythON', 'pYthON', 'PYthON', 'pYTHON', 'PYTHON']
+        word_list += ['search', 'Search', 'sEARCh', 'SEARCh', 'searcH',
+                      'searCH', 'SearCH', 'sEarCH', 'seARCH', 'SEARCH']
+
+        for word in word_list:
+            res = self.testapp.get(
+                u'/api/v1/admin/bmarks/{0}?&api_key={1}'.format(
+                    word, API_KEY),
+                status=200)
+            data = json.loads(res.body)
+
+            self.assertTrue(
+                data['count'] > 0,
+                "Should have found > 0 bookmarks")
+
+        self._check_cors_headers(res)
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -28,3 +28,4 @@ Thanks to these people that have contributed code/assistance to bookie
 - Jelmer Vernooij
 - Kyle Adams
 - Surya Pappoppula
+- Yash Mehrotra
--- a/bookie/models/__init__.py
+++ b/bookie/models/__init__.py
@@ -441,6 +441,7 @@ class BmarkMgr(object):
                 from_self()
 
         if tags:
+            tags = [tag.lower() for tag in tags] # For case matching
             qry = qry.join(Bmark.tags).\
                 options(contains_eager(Bmark.tags))
 


craigmaloney:
Hi @yashmehrotra,

Thank you for this patch. Would you add a test case that checks that the tags are indeed matched? Ideally it would test upper, mixed, and lower case. It should fail without your patch applied.

Thanks again!

yashmehrotra:
@craigmaloney Ok, I am adding them currenty , but I may need some help. At what time are you available on the IRC channel?

craigmaloney:
@yashmehrotra I'm at work most of the day (EST) but should be available for a bit off-and-on.

yashmehrotra:
Ok thank you

yashmehrotra:
@craigmaloney I did what you said, I added tests and without my patch the test fails (and succeeds with the patch) . Kindly review my code.
Thanks for the help.

bookie/tests/test_api/test_base_api.py
```diff
@@ -1294,3 +1294,40 @@ def test_user_stats(self):
         self.assertTrue(
             'with_bookmarks' in data,
             "Should have count of users with bookmarks: " + str(data))
+            
+    def test_tag_search_casing(self):
+        """Test that search works same with all types of casing"""
+
+        # Adding a tag
```
mitechie: You shouldn't need to fetch the api key. There are helpers for this in place already. 

https://github.com/bookieio/Bookie/blob/develop/bookie/tests/test_api/test_base_api.py#L44

bookie/tests/test_api/test_base_api.py
```diff
@@ -1294,3 +1294,40 @@ def test_user_stats(self):
...
+        res = DBSession.execute(
+            "SELECT api_key FROM users WHERE username = 'admin'").fetchone()
+        key = res['api_key']
+
+        test_bmark = {
```
mitechie: there's a helper to add a bookmark for you as well https://github.com/bookieio/Bookie/blob/develop/bookie/tests/test_api/test_base_api.py#L60

bookie/tests/test_api/test_base_api.py
```diff
@@ -1294,3 +1294,40 @@ def test_user_stats(self):
...
+            'username': u'admin',
+            'is_private': 'true',
+        }
+
+        res = self.testapp.post('/api/v1/admin/bmark',
+                                content_type='application/json',
+                                params=json.dumps(test_bmark),
+                                status=200)
+
+        # Testing the same tag 'bookie' with different cases
```
mitechie: so this is the meat of this test.

bookie/tests/test_api/test_base_api.py
```diff
@@ -1294,3 +1294,40 @@ def test_user_stats(self):
...
+        word_list = ['bookie', 'Bookie', 'bOokie', 'BOokie', 'boOkie', 'BoOkie', 'booKiE', 'BooKiE', 'BOokIE', 'boOkIE', 'BoOkIE', 'bOOkIE', 'BOOkIE','BOOKIE']
+
+        for word in word_list:
+            res = self.testapp.get(u'/api/v1/admin/bmarks/{0}?&api_key={1}'.format(word, API_KEY),
+                status=200)
+            data = json.loads(res.body)
+            
+            self.assertTrue(
+                data['count'] > 0,
+                "Should have non-zero tags count")
```
mitechie: this is more "Should have found > 0 bookmarks" not really a tag count correct?

bookie/views/api.py
```diff
@@ -416,6 +416,9 @@ def bmark_recent(request, with_content=False):
     # string in a query string
     if not tags and 'tag_filter' in params:
         tags = params.get('tag_filter').split()
+        
+    if tags:
+        tags = [tag.lower() for tag in tags]
```
mitechie: so this is cool here, however this I think should live in the BmarkMgr.find() method. So that every place that find method is called the lower casing is done on the model end vs the view end. It's a bit of a case of "fix it in the one place everyone uses it" vs this one case.

mitechie:
Thanks for the update! This looks really good. I think the test can be simplified and the code required there cut in half. 

I also think that the better place to lowercase is in the BmarkMgr vs the view itself. Let me know what you think after looking into that a bit.

yashmehrotra:
@mitechie Ok , I will make the relevant changes and put the case changing in BmarkMgr.find()

yashmehrotra:
@mitechie Updated everything as you said . Thanks for the guidance.

craigmaloney:
:+1:

yashmehrotra:
@craigmaloney Thanks , so is it good for a merge?

yashmehrotra:
@mitechie Converted it all into 1 commit .

PR #531, opened 2014-11-22
Minor Lint

yashmehrotra:
@mitechie Hi, I am really sorry. I added a comment after 'make lint' was clean (in the previous commit). I hadn't checked it that time but the 'make lint' gave me a whitespace error due to my 'one spaced inline comment'
Here is that minor fix.
These mistakes won't happen again. Sorry.


mitechie:
ty for the fix

Issue #532, opened 2014-11-29
parsed version not available through API?


anarcat:
hi

i have created a bookmark through the API using the following call:

```
http --json POST  https://example.com/bookie/api/v1/anarcat/bmark?api_key=CENSORED url=http://koumbit.org/ tags=test description=koumbit
```

(this is using https://github.com/jakubroztocil/httpie)

unfortunately, the resulting bookmark doesn't have a rendered HTML version. is that normal? is that rendered server-side? the [API docs](https://github.com/bookieio/Bookie/blob/develop/docs/api/user.rst) are not quite clear about this...

mitechie:
Thanks for the bug @anarcat. This is expected at first. There is a celery job that goes back over bookmarks without readable content and attempts to fetch it. It runs nightly I believe. It then adds the readable view and content at that time. If you have the page ready, you can pass the content to the api call so that it's there right away. This is how the extensions work. They grab the current HTML on the page and send it with the request to save the bookmark to the API. 

If you want, check out the celery task https://github.com/bookieio/Bookie/blob/develop/bookie/bcelery/tasks.py#L300 and if you bump the time on that back to a short timespan and manually restart celery you should see it go out and load the content.

mitechie:
Closing as expected behaviour, let me know if the explanation doesn't make sense or I've forgotten how things are put together and feel free to reopen it.

anarcat:
thanks for the feedback!

maybe the API docs should be clarified so that it's clearly stated that the HTML page can be provided through the bmark submission... is it the "rendered" version or the "original" version that should be submitted to the API? ie. does the extension do the processing or does the API server side render?

PR #533, opened 2014-11-29
clarify what the "content" field means

anarcat:
this is related to #532, and should make it clear that the `content` field needs to be provided to have a prompt readable version available, and what happens if the field is not provided.


mitechie:
Thanks you for the update.

PR #536, opened 2014-11-30
mention the willie bot extension

anarcat:


mitechie:
Thanks, this looks ok but I think it's in markdown instead of rst format so the links don't look right when the page is generated. Can you double check the format please?

anarcat:
done, sorry for the noise.

mitechie:
ty for the update!

PR #538, opened 2014-12-04
Support hosting under a domain subpath (eg. example.org/bookie)

anarcat:
This adds support for hosting Bookie on a subdomain. Some things still
don't work quite right (ie. sometimes the slash after `bookie` in
`example.org.bookie` is lost) but that's not frequent enough to be a
blocker in my opinion.

We have been using this in production for months without much problem
except the above now.

### DIFF
--- a/bookie/templates/accounts/index.mako
+++ b/bookie/templates/accounts/index.mako
@@ -3,6 +3,8 @@
 <%namespace file="func.mako" import="account_nav, password_reset"/>
 <%
     date_fmt = "%m/%d/%Y"
+    app_url = request.route_url('home').rstrip('/')
+    app_path = request.route_path('home').rstrip('/')
 %>
 ${account_nav()}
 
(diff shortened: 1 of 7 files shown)


anarcat:
any comments here?

craigmaloney:
Apologies, we're a little busy over on this side. We'll take a peek at this shortly.

Thanks!

mitechie:
This looks fine. I'd like to see what's up with the missing slash and try it for a bit before landing. A test setup would be good but might be a bit crazy to check everywhere a url is generated.

mitechie:
Thanks for this PR. I'm landing this to get it into trunk and all tests pass. It'd be great if you could setup some docs on your install/setup for the subdirectory to help users take advantage of the functionality.

anarcat:
where should that doc live? in https://github.com/bookieio/Bookie/blob/develop/docs/hosting.rst?

mitechie:
I think that would be perfect.

Issue #541, opened 2014-12-09
allow hosting google fonts locally


anarcat:
while I can appreciate nice fonts, i think it's a problem that bookie refers its users to Google-hosted fonts. google has and collects a lot of information about its users, through all those free services, and having Bookie an independent service would require those fonts to be hosted elsewhere.

mitechie:
Thanks I'm all for this as an option. We do it with the extensions I believe. I'm tagging this as a bite size bug someone new could work on.

NalinG:
@shahhena95  shouldn't we use .ttf font type rather than .woff2 , since it is not supported by some browsers.

NalinG:
I can give a PR regarding the same if you want.

anarcat:
@mitechie have you looked at 23717a5 from @shahhena95? it looks like a simple enough fix.. i wonder if the font's license is compatible however.

anarcat:
ping

PR #544, opened 2014-12-24
[issue #541] hosted google fonts locally

shahhena95:
Now fonts are called from static folder instead of google URL.


mitechie:
We use both Cabin and Cabin Sketch in the site:

```
=Cabin|Cabin+Sketch:bold&v2
```

I also think this should be configurable. For the public site, it's a nice performance boost since the fonts come from the Google CDN and because it's a different domain most browsers will parallel this download along side ones from bookie.io as well. 

http://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser

I'd prefer if this took a config option in the ini file and used that to determine if it should use local or remote urls for the fonts. This would require some logic then in the template based on that config.

shahhena95:
@mitechie Google sends different format according to the user agent. Like Firefox gets .woff format and Chrome gets .woff2 format .. Can you tell me what to do?

mitechie:
Looking at the css rules I think the trick is that we need to rely on the fontawesome.css. 

If you look in the extensions where we use it locally the rule cascades the font file used like:

https://github.com/bookieio/bookie-chrome/blob/master/chrome_ext/fontawesome/css/font-awesome.css

```
  src: url('../fonts/fontawesome-webfont.eot?#iefix&v=4.1.0') format('embedded-opentype'), url('../fonts/fontawesome-webfont.woff?v=4.1.0') format('woff'), url('../fonts/fontawesome-webfont.ttf?v=4.1.0') format('truetype'), url('../fonts/fontawesome-webfont.svg?v=4.1.0#fontawesomeregular') format('svg');
```

PR #545, opened 2014-12-29
fixed login-redirect issue for authenticated user

AnishShah:
After a user is successfully logged in, and the user hits "https://bookie.io/login", then the user is not redirected to its recent page.
This pull request fixes this issue.

### DIFF
--- a/bookie/tests/test_auth/test_api_base.py
+++ b/bookie/tests/test_auth/test_api_base.py
@@ -124,9 +124,9 @@ class TestAuthWeb(TestCase):
             'Failed login' in str(res),
             "Should have 'Failed login' in the resp: " + str(res))
 
-    def test_login_after_successful_login(self):
-        """Verify it redirects to the user's /recent after
-        succesful login"""
+    def test_redirect_for_authenticated_user(self):
+        """Verify if authenticated user gets redirect when
+        it hits login URL"""
 
         user_data = {
             'login': u'admin',
--- a/bookie/tests/test_auth/test_api_base.py
+++ b/bookie/tests/test_auth/test_api_base.py
@@ -123,3 +123,28 @@ class TestAuthWeb(TestCase):
         self.assertTrue(
             'Failed login' in str(res),
             "Should have 'Failed login' in the resp: " + str(res))
+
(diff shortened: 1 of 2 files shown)


bookie/views/auth.py
```diff
@@ -82,6 +83,12 @@ def login(request):
             message = "Failed login"
             AuthLog.login(login, False, password=password)
 
+    if isinstance(request.user, User):
```
mitechie: I'd like this to be a bit smarter. What about not just checking there is a User, but that it's a current valid user in the system that is activated as a security measure?

bookie/views/auth.py
```diff
@@ -14,6 +14,7 @@
 from bookie.models import IntegrityError
 from bookie.models.auth import UserMgr
 from bookie.models.auth import ActivationMgr
+from bookie.models.auth import User
```
mitechie: I know it's not your doing but we've been trying to clean up the imports more into the format like:

```
from bookie.models.auth import (
    ActivationMgr,
    User,
    UserMgr,
)
```

bookie/tests/test_auth/test_api_base.py
```diff
@@ -123,3 +123,28 @@ def test_login_null(self):
         self.assertTrue(
             'Failed login' in str(res),
             "Should have 'Failed login' in the resp: " + str(res))
+
+    def test_redirect_for_authenticated_user(self):
```
mitechie, line 127: given the updated sanity check below those should have matching tests to make sure we verify we force a login to invalid users.

mitechie:
Thanks for this. I'd like to push on a little more awesomeness for this to land if that's ok.

bookie/tests/test_auth/test_api_base.py
AnishShah, line 127: `test_login_url` test on Line26 already does that. That's why, I didn't add another test case.

AnishShah:
@mitechie I have made the changes. 
If you want a separate test case for invalid user (which `test_login_url` already does) , then I'll do it.

mitechie:
Thanks for the update!

PR #546, opened 2014-12-30
GitHub API integration

AnishShah:
I have integrated Github API in Bookie. I have not used any package for Github API since most of them that I came across use Basic Auth instead of OAuth. I have checked most of the errors that can occur. Please have a look.


craigmaloney:
Hi Anish,

It looks like there's a merge conflict. Please check to see if you can resolve this and resubmit the push request. If you have then please delete this request.

Thanks!

Issue #551, opened 2015-01-09
bookie url click doesn't work if the url contains extended chars


alexlehm:
I found a few urls I had in my import from delicious that contained an accented character in the url, e.g. amazon urls that contain the book title in the url. This doesn't happen when using an url copied from the browser url field since that is encoded, but I could reproduce it by manually entering an url.

When clicking on an url of that type in the bookie list, the redirect url gives a 404 error and editing the url does as well.

To reproduce this, click on add bookmark and enter this url http://www.amazon.de/Molwanien-schadhaften-Lächelns-Santo-Cilauro/dp/3453811380 , e.g. in my bookmark list, this gives the following link, which doesn't work https://bookie.io/alexlehm/redirect/a3aab825f7d7ef

Its possible to get to the link by copy'n'paste, but it would be better if the urls were converted or encoded to work correctly.

mitechie:
Thanks for the report and sorry for the trouble. I though we had our utf8 in order. We'll peek at it.

alexlehm:
of course it could be argued that this is not a valid url, since a browser will encode this as %C3%A4 automatically. when adding the url with the encoded chars, everything works

AnishShah:
@mitechie working on this..

PR #552, opened 2015-01-15
[issue 551] fixed issue of url containing extended chars

AnishShah:
This solves the issue #551.
But I'm not able to write testcase for this issue. 

When I add something like this ->

```
test_url = u'http://www.amazon.de/Molwanien-schadhaften-Lächelns-Santo-Cilauro/dp/3453811380 '
```

I get -> `SyntaxError: Non-ASCII character '\xc3' in the file`

I added `# -*- coding: utf-8 -*-` at the top of the file. But then, it passes the testcase without any changes in the original code.

Please help me in adding the testcase. ?


preddy5:
Hai @AnishShah, one reason could be you copy pasted the url and the unicode char are not visible on your text editor, once try by typing the url instead of pasting

PR #553, opened 2015-01-17
fixed normalizing of URL containing Non-ASCII chars

AnishShah:


bookie/tests/test_webviews/test_webviews.py
```diff
@@ -143,6 +144,33 @@ def test_existing_url_entry_error(self):
             })
         self.assertIn(existing_url_message, res.body)
 
+    def test_bookmark_url_encoding(self):
+        """Verify that the URL containing Non-ASCII chars are normalized"""
+        self._login_admin()
+
+        test_url = u"http://www.amazon.de/Molwanien-schadhaften-" \
+                   u"Lächelns-Santo-Cilauro/dp/3453811380".encode('UTF-8')
```
craigmaloney: Not sure but I think this is double-encoding.

AnishShah: @craigmaloney  Unicode is behaving very strange. It replaces `ä` by `?`. Do you know why? String works properly.

AnishShah:
@craigmaloney Unicode is behaving very strange. It replaces `ä` by `?`. Do you know why? String works properly. So, I have changed the `test_url` to string, instead of unicode.

craigmaloney:
Thank you!

PR #564, opened 2015-06-17
Feature/authed starttls email out

adamlincoln:
Allows outgoing signup emails to be sent with starttls and user/pass authentication for smtp.

Also fix signup logging.

### DIFF
--- a/bookie/lib/message.py
+++ b/bookie/lib/message.py
@@ -92,11 +92,20 @@ class Message(object):
             try:
                 all_emails = msg['To']
                 smtp_server = self.settings.get('email.host')
+                starttls = self.settings.get('email.starttls', False)
+                smtp_user = self.settings.get('email.smtp_user', None)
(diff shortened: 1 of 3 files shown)


bookie/lib/message.py
```diff
@@ -92,11 +92,20 @@ def send(self, message_data=None):
...
 
                 if smtp_server == 'sendmail':
                     sendmail(msg['To'], msg['From'], msg['Subject'], self.body)
                 else:
                     mail_server = smtplib.SMTP(smtp_server)
+                    if starttls:
+                        mail_server.starttls()
+
+                    if smtp_user is not None and smtp_password is not None:
+                        mail_server.login(smtp_user, smtp_password)
```
adamlincoln, line 107: This raises SMTPAuthenticationError if the credentials are bad, but I couldn't immediately figure out how to log that from here (SignupLog() didn't seem to behave the way I expected).

